PR #49, opened 2020-03-19
updated hospital beds data for CZ + healthcare system capacity

vejmelkam:
Hi,

I updated the hospital beds data in 'assets/data/population.json' based on current per-capita counts reported recently here [source in Czech]:

https://www.irozhlas.cz/zpravy-domov/koronavirus-evropa-data-srovnani-jednotky-intenzivni-pece-jip-plicni-ventilatory_2003190630_nkr

Please let me know if that's the right place to modify the data, I grepped the codebase to see if it was generated but it doesn't seem to be.


rneher:
Thanks. Unfortunately, these data are generated. We just moved the data to a separate repo. Would you mind making this change here:

https://github.com/neherlab/covid19_scenarios_data/blob/master/populationData.tsv

Sorry about the duplicated effort. Will add a note to the README.

vejmelkam:
No problem, will do.

PR #55, opened 2020-03-20
Fix navbar home link

rcbevans:
Replace navbar brand logo href with Link pointing to the root /covid19 page.

The previous behavior navigated the app to http://localhost:3000/, which gives a "Cannot GET /" error
since webpack is serving the root page from http://localhost:3000/covid19/.


rcbevans:
It looks like someone else already checked in a change updating the webpack output route to serve from /, so this change is no longer required.

Issue #19, opened 2020-03-17
Add tests


ivan-aksamentov:
In order to perform refactoring and to add new features with more confidence, we need unit, integration and end-to-end tests for:

 * algorithm parts

 * React parts - components and pages

 * Redux parts - state management: reducers and sagas

The infrastructure:

 * `jest` and `@testing-library`are already set up. Ready to run some tests

 * `cypress`is planned for e2e tests

 * We plan either Travis CI or CircleCI for continuous integration

Any help is appreciated!

Joluma:
@ivan-aksamentov did you already start implementing e2e with cypress? I can help with that.

ivan-aksamentov:
@Joluma 
Hi Joël, I haven't started yet. But the e2e testing is probably one of the most valuable things to contribute.

It may be tricky, because many parts of the project are changing rapidly. For example, I have a huge Redux revamp coming soon(ish). So the test will have to be adjusted frequently.

Feel free to try it out!

ivan-aksamentov:
Some more thoughts for unit and integration tests:

 - The goal is to bring confidence that refactors and new features don't break the results.

 - Most importantly, care should be taken to not produce wrong science in the end! That's people's lives counts, not just numbers we are juggling there.

 - Focusing on testing `run()` function is somewhat pointless, I think. It's an entry point of the algorithm and it has too many parameter combinations and too many possible results to cover. 

 - Instead, we should focus on writing slightly more granular tests for everything in the `src/algorithms/` directory.

 - Overall, in terms of thesting, I am in the same camp as Guillermo Rauch (@rauchg) and Kent C. Dodds (@kentcdodds): "Write tests. Not too many. Mostly integration." .

Adding more tests and improving the app testability will ensure that we can expand the app's usability and add more interesting algorithmic features.

Contributions are desperately needed!

szonjajakab:
I could do the CI building if its not taken.

owais-eng:
how can i get the code of this projectSent from my Samsung Galaxy smartphone.
-------- Original message --------From: szonjajakab <notifications@github.com> Date: 21/03/2020  2:55 pm  (GMT+05:30) To: neherlab/covid19_scenarios <covid19_scenarios@noreply.github.com> Cc: Subscribed <subscribed@noreply.github.com> Subject: Re: [neherlab/covid19_scenarios] Add tests (#19) 
I could do the CI building if its not taken.

[...]

PR #60, opened 2020-03-20
#19 Fix silent failure of 'should work for a lot of countries'.

abrie:
This PR is part of a larger issue, described by #19.

It addresses a specific test named "should work for a lot of countries". It was failing because it used the wrong Country Name key to look up Population Scenario and Country Age Distribution. The failure was masked because the test was wrapped in a try/catch.

This PR fixes the test and removes try/catch.

### DIFF
--- a/src/algorithms/run.test.tsx
+++ b/src/algorithms/run.test.tsx
@@ -78,22 +78,24 @@ describe('run()', () => {
     const results: Record<string, any> = {}
 
     for (let country of countries) {
-      try {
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
Hi @abrie thanks for the contribution!

Our tests are a bit of a joke right now as you see. I hope this makes it a bit better.

If you have some more time to continue, I added some more thoughts here: 
https://github.com/neherlab/covid19_scenarios/issues/19#issuecomment-601984057

Issue #56, opened 2020-03-20
Internationalization


mserranom:
It would be really useful to support translations to other languages.

https://react.i18next.com/ is a reasonable option given the current tech stack

n0f3:
what languages are you thinking?

mserranom:
spanish at least

rapzo:
I might pick this one up. And I can do the portuguese version, since we definitely need help as well!

lukepighetti:
English & Spanish at a bare minimum, imo.

charlie632:
I can help translating to spanis and/or helping with i18Next. Let me know if you are working on this @rapzo

rapzo:
Doing the setup right now. Once i get it working i'll push to my fork and will open to whomever willing to stretch a hand!

charlie632:
Great!

ErikSchierboom:
I can help with the Dutch translation!

gcollard:
I'll help with the French translation

samarv:
I can help with hindi!

goktugerce:
I can help with the Turkish translation.

t0byman:
> I can help with the Dutch translation!

Kerel! Hulp nodig!?

gleiderr:
I can help with portuguese

Nebadon1:
Spanish if you need more people

ivan-aksamentov:
That's fantastic, folks!

I've merged this PR from @rapzo 
https://github.com/neherlab/covid19_scenarios/pull/62/

TODO:
 - language switch
 - scaffold new locale files
 - fill in the new locale files

Cant wait to see this goes forward!

besrabasant:
I will too help in Hindi.

FireLizard:
I can help with the German translation.

hannesgr:
I can help with the Swedish translation

rapzo:
I'm gonna work on extracting all strings from components. It's a long one... But once the english is complete it will be easier to add new stuff. But if anyone is already at it please open a WIP PR so we don't overlap each other.

hannesgr:
I've just started, so I'll add a WIP PR.

hannesgr:
#76

Emmanuel-R8:
> I'll help with the French translation

Ditto

hannesgr:
All right, I've tried to identify all translatable strings. 

I made a complete PR Here #85. Didn't translate the About page though, not sure about how to go about that one?

[...]

PR #62, opened 2020-03-20
feat(#56): add i18m support with i18next and react-i18next

rapzo:
Rough implementation... Lost a bit of time around on how the whole setup is wired in order to have lazy load, but stopped, kept it simple, and open PR.


ivan-aksamentov:
Thanks @rapzo 
This is fantastic! Looking forward to the developments in this area!

PR #64, opened 2020-03-20
Prevent having button inside button on CollapsibleCard-CardHeader

ivnnv:
Not a big change but its the first thing I realized when running the project, the console complained about html validation as <button> can not contain <button> inside.

button inside button

This PR solves that while keeping the same design and functionality at the CollapsibleCard CardHeader


ivan-aksamentov:
Hi @ivnnv ,

Thanks so much for fixing this! Our team hated this pesky warning but was busy with something else.
Please contribute again! :)

PR #70, opened 2020-03-21
internationalization - Language Switcher

nonotest:
## Description

This adds a basic language switcher in the header and the files for french language.
The selected language is stored in localstorage.
We could add flags in the display but I didn't want to pull an extra dep just like this.

## Related issues

https://github.com/neherlab/covid19_scenarios/issues/56

internationalization

## Testing

Navigate to the only page using i18n - Team
Change the language, see the change
Reload the page and check that the change was persisted.


rapzo:
Awesome!
Let's go!

nonotest:
small gotchas for later
1) A bare dropdown won't scale with too many elements / is not good on mobile (select box) 
2) we may not want this in the header when we have more nav links
3) a solution with the locale in the url path may be better suited long term

rapzo:
There is a TODO covering those issues: https://github.com/neherlab/covid19_scenarios/issues/56#issuecomment-601990919
Fiddling with routes, language discovery and cached version are issues to freeze a good week of work so, maybe keeping it small, yet incomplete, may be the best approach, as long as stuff keeps coming in, imho.

ivan-aksamentov:
@nonotest @rapzo 
I like the PR.

Dropdown is fine for now. There will not be a lot of links.
On mobile we should figure out something with navbar responsively anyways. Maybe collapse into a burger menu. and the switcher will be different too.  URL param does not exclude the switcher IMO.

Feel free to continue the discussion or open more PRs if you want to address any of it differently.

PR #73, opened 2020-03-21
feature: add e2e tests capabilities with cypress, add one first test for the navigation

Joluma:
## Description

This PR is setting the ground for the E2E tests using Cypress:
- installing Cypress
- setting typescript support for the tests
- add one test, for the navigation bar

## Related issues

- Contributes to https://github.com/neherlab/covid19_scenarios/issues/19

Will affect testing area

## Testing

`yarn run e2e` will run the tests in a headless browser, the result will be seen as below:
e2e-tests

(The test run is saved as a video and the result can be seen in the `/cypress/videos/` folder)

`yarn run e2e:dev` will run the tests in a visual browser (electron, chrome and firefox dev available at the moment), the result will be seen as below:
e2e-tests

## Deploy Notes

Cypress is using the environment variable set in the `cypress.json` file. During the CI/CD if we want to run the tests against another environment we need to set this variables.
Right now, the only needed variable is `CYPRESS_BASE_URL`.


cypress/integration/navigation.spec.ts
```diff
@@ -0,0 +1,29 @@
+/// <reference types="cypress" />
+
+import links from '../../src/links'
+
+context('The navigation bar', () => {
```
kibertoad: What's the longer-term plan to handle navigation? Maybe introducing something like PageObject pattern would be helpful?

ivan-aksamentov:
@Joluma Great work! Would you be able to continue on adding more tests?

Joluma:
@ivan-aksamentov Hi, thanks. Yes, I'll try to add more piece by piece. Do you have some exact requirements for the tests? Or some user scenario you want to cover in priority?

ivan-aksamentov:
@Joluma Honestly, I did not think much about e2e scenarios yet.
I added some thoughts in the new issue here: https://github.com/neherlab/covid19_scenarios/issues/119

PR #75, opened 2020-03-21
feat(Test): Initial tests for the infectionRate model

kibertoad:
ref #19

This PR introduces following changes:

1) Model directory is cleaned up to provide better visibility regarding responsibilities of different files (what it types, what are utils, what is the actual model). This makes directory less cluttered and makes it easier to understand what is the surface in need of testing.

2) Added JSDoc for model under test; not sure if there are plans to convert it to TS in the future, meanwhile JSDoc is helpful for explaining the data at play;

3) Initial set of tests for infectionRate() added. Not sure what the most maintainable approach for checking values would be here - I've opted for in-line asserted values, but jest snapshots might be a better option, if they won't be abused by blindly regenerating everything.


nnoll:
This looks great -- a much needed refactoring. I don't see any changes to the model so it looks good from my end. I'll tag Ivan to have final say re: testing but I think this should be merged.

ivan-aksamentov:
@kibertoad  This is fantastic! Exactly what we need. Merging right away.

A few notes:

 - Regarding  JSDocs, as algorithm evolves, inline docs may get somewhat dated. 
 - Typescript for `model.js` would be fantastic! It's a large hole in our type system.
 - I think inline values are fine. We want people to fix tests on changes.
 - I am a bit concerned that the equality checks would work correctly at all times with floating point numbers. Should we use`.toBeCloseTo()`instead?

PR #80, opened 2020-03-21
Add Favicon, Increase size of input type number

bharath6365:
## Description
1) Add favicon for the application. (Uses the same as that of the parent domain)
2) Increase the size of input type number thereby giving users more room to click.
number-size-increase

 - Favicon
 - Forms


bharath6365:
There are some UI Breakages on this one. Will reopen shortly.

PR #82, opened 2020-03-21
fix: do not re-render Mitigration graph unnecessarily

### LINKED ISSUE #72: Fix black dots dragging behavior on the mitigation plot (React hooks + D3.js guru needed)
## Steps To Reproduce

1. Refresh the main page
2. Mouse-focus into one of the form fields, for example increment "Imports per day" number.
3. Without blurring (moving the cursor out of the form field), try to drag a black dot on the "Mitigation" plot

## The current behavior

The "Mitigation" plot dot being dragged is not moving until mouse button is up

## The expected behavior

The "Mitigation" plot dot being dragged is not moving as the cursor moves

## Hypotheses

The blur event out of the field coincides with the initial click on the plot's dot. After the blur, the form rerenders and also causes the plot to rerenders.
Currently there is a hack. This `.html(null)` on every render clears the entire plot:

https://github.com/neherlab/covid19_scenarios/blob/b07eeaedd82dd3f54067e241cb62e5e2ac879504/src/components/Main/Containment/ContainmentGraph.tsx#L44-L49

This also causes the destruction of the  black dot we are dragging. The new dot will be rendered only on mouse up. Meanwhile we are dragging nothing.

## Potential solution

We need to remove the hack shown above and to figure out proper dependencies for the `useEffect hook here:

https://github.com/neherlab/covid19_scenarios/blob/b07eeaedd82dd3f54067e241cb62e5e2ac879504/src/components/Main/Containment/ContainmentGraph.tsx#L225-L230

such that the plot is properly cleared when it is needed, that it is responsive (to screen size) and not causing unnecessary rerenders at the same time

D3-React interop in this component should be properly audited for additional bugs and performance problems.

petebacondarwin:
Previously the graph was being re-rendered on every
event, which caused problems when trying to drag a
graph vertex at the same time as de-focussing another
input element.

Now the graph is only rendered initially, and then when
either the graph data changes, or the graph is resized.

Fixes #72

### DIFF
--- a/src/components/Main/Containment/ContainmentGraph.tsx
+++ b/src/components/Main/Containment/ContainmentGraph.tsx
@@ -17,6 +17,9 @@ export interface DrawParams {
 }
 
 function draw({ data, width, height, onDataChange, d3ref }: DrawParams) {
+  if (d3ref.current === null) {
+    return
+  }
   const newData = data.map(d => {
     return { t: d.t, y: d.y }
   }) // copy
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
@petebacondarwin Great investigation Pete!
I was struggling with this thing so much and you just fixed it in 2 lines of code basically.
We are very happy to have you in our club :)

Issue #71, opened 2020-03-21
Improve mobile user experience


ivan-aksamentov:
Our current mobile experience is suboptimal. 

### Key defects:
 - Situation is particularly bad with charts and plots, which is a central aspect of the application.
 - Some of the elements are needlessly large, some are to small
 - Scrolling is awkward
 - Tapping "Run" button gives no clue that there is more content appeared below
 - Calendar is cut-off

We currently use reactstrap as a frontend library with some Sass stylesheets and bootstrap utility classes sprinkled here and there. As a team of 3 people, we would not able to change that. But if we find knowledgeable contributor(s) in UI/UX space, then the transition to another styling solution will become more realistic, including the full revamp of the UI.

Our main focus is desktop, with 720p or 1080p screen at least, but having broken mobile experience hurts the outreach, because we mostly promote our project on social media (Twitter). People are coming from there on the phones first, and they are probably not very impressed and are less compelled to switch or return with a larger device.

### Related issues:

 - Improve desktop user experience: https://github.com/neherlab/covid19_scenarios/issues/13

Screencast of "iPhone X" simulation in Chrome:

![mobile](https://user-images.githubusercontent.com/9403403/77221789-3986a280-6b4d-11ea-99fd-2d3b1eef4ec9.gif)

ivan-aksamentov:
This is a huge issue. Feel free to open separate issues for smaller fixes and mention this one.

miksa1987:
Hi! I could start working on the elements. As for clicking run and more items appearing, could it be done so that form items would be kind of like menu overlay on top of everything, and when clicking run, they go away to a menu button ready to be invoked again? 
Also, I was thinking about the popups in graphs, perhaps they should be placed on top of the graph on mobile, so that it wouldn't be on top of the pointer making it hard to tap on certain spot?

ivan-aksamentov:
Hi @miksa1987,

> I could start working on the elements
It'd be awesome! Make sure you coordinate with other folks, such as in #83  and other issues and PRs listed above

> form items would be kind of like menu overlay on top of everything, and when clicking run, they go away to a menu button ready to be invoked again
That's a great design idea! 

> about the popups in graphs, perhaps they should be placed on top of the graph on mobile
That'd be fantastic! Although, I am not sure if our graph library allows that to be done easily or even at all. But I'd be glad if you can look into it!


[...]

PR #87, opened 2020-03-21
Fix Calendar Alignment on Mobile

bharath6365:
## Description
Fixed the logic that decided the number of months that had to be rendered on mobile. Previously it was rendering 4. We have space only to render 1.

## Related issues
https://github.com/neherlab/covid19_scenarios/issues/71

Date Picker

### DIFF
--- a/src/components/Form/FormDatePicker.tsx
+++ b/src/components/Form/FormDatePicker.tsx
@@ -10,6 +10,15 @@ import FormLabel from './FormLabel'
 
 import { DateRange } from '../../algorithms/Param.types'
 
+
+// Function to determine number of months to display on the datepicker.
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
Thanks @bharath6365!

Could you please also investigate how it looks on iPad emulator in Chrome?
In particular, if `withPortal` option of  `DateRangePicker`. For some reason portal rendering is actiaeted on iPad, but not in iPhone X, for example.

Could we avoid hardcoding number of months at all?

I found this storybook to be useful to explore the options:
https://airbnb.io/react-dates

Don't hesitate also if you think that there is a better option than `react-dates` we can replace it then.

Issue #17, opened 2020-03-17
IMPORTANT: Autorun: run the algorithm and update the plots automatically without clicking on Run button


ivan-aksamentov:
This may include:

  - debounce `run()` calls
  - hide feature behind the checkbox, disabled by default

Issues:

 * `formik` does not allow to subscribe to the form state changes. It requires submit to be called
 * `formik` does not expose the form state. We currently grab the state in validation callback, which arguably, is a hack
 * `formik` seem to be re-rendering the entire form on changing any field. This is a performance problem - In automatic mode editing fields is very uncomfortable.  There are ways to workaround it, but I don't want to introduce any more workarounds.

Solutions:
 - Replace `formik`  with `redux-form`? It will require quite substantial changes, for example reimplementing the validation.

I posted a cry for help in the Formik's chat:
https://spectrum.chat/palmer/formik/retrieve-values-without-submit-re-rendering-covid-19-scientists-need-help~d595a68c-2090-4a16-8a21-6f0772ff48e7

The following gif shows how on changing one value the entire form including mitigation plot is being re-rendered:

![Peek 2020-03-16 20-55 mp4](https://user-images.githubusercontent.com/9403403/76856275-3b055180-6853-11ea-8194-fd8c0ed0518f.gif)

UnforbiddenYet:
@ivan-aksamentov you may check out [the Fokmik's author answer](https://github.com/jaredpalmer/formik/issues/1218#issuecomment-500225537) to the same question where he suggests the possible solution for auto form submission.

As for the optimization, I'd say using the [`<FastField/>`](https://jaredpalmer.com/formik/docs/api/fastfield) or custom [hook-based](https://github.com/jaredpalmer/formik/issues/1026#issuecomment-524594409) implementation should prevent form fields wasted renders.

ivan-aksamentov:
@UnforbiddenYet 
Hi Dmytro,
Thanks so much for the info. I will look into the Jared's example and the `FastField`.

benjwexler:
Hey @ivan-aksamentov, Formik does expose the form’s values as part of the `formikBag` in the render method. Why not add an `onChange` to the Form component like this?

          {({ values, errors, touched, isValid }) => {
            const canRun = isValid && severityTableIsValid(severity)
            return (
              <Form
                onChange={(ev) => {
                  const { name, value } = ev.target;
                  const newValue = {};
                  newValue[`${name}`] = value;
[...]

PR #90, opened 2020-03-21
feat: add autorun toggle

3b3ziz:
## Description
This pull request implements the autorun feature based on the values object provided by formik.

## Related issues
Fixes https://github.com/neherlab/covid19_scenarios/issues/17


ivan-aksamentov:
Hi @3b3ziz thanks for the efforts!

Unfortunately this makes the for too slow. For example, if I go to the deployed version:
https://covid19-scenarios-git-fork-3b3ziz-feat-autorun-onchange.covid19-scenarios.now.sh/
and try to increase the population with arrows or to type in any numeric field, the entire apo hangs for a second or two after every click or key stroke. We cannot allow this kind of user experience.

I have a branch, where I tried to implement this and also revamped the whole state and state updates in Redux. Bit it is unfinished and it's a bit of a mess, but maybe you'll find something useful useful.
https://github.com/neherlab/covid19_scenarios/compare/feat/autorun

Another minor thing is that we want to autorun the algorithm on app startup as well.

I hope you can fix the freezes. That would be fantastic!

3b3ziz:
@ivan-aksamentov Could you please check again the speed of the deployed version now? https://covid19-scenarios-git-fork-3b3ziz-feat-autorun-onchange.covid19-scenarios.now.sh/

I have used a different (Supposedly more efficient) approach and also added the autorun to work on app startup.

Generally, I think there is a performance problem since I noticed that the app renders multiple times when changing input fields. I will try to investigate in that as well.

ivan-aksamentov:
@3b3ziz I did not notice any changes, it still freezes. 
Repro: try to delete the number in "Populations" field and then quickly (within 1 second) to enter "123456789". It takes  about 5-7 seconds before the UI get responsive again. I suspect that the algorithm runs 10 times or so, after every keystroke.

ivan-aksamentov:
@3b3ziz it is a hard problem with the current setup. That is why I haven't solved it yet. And nobody else seem to be taking it. I am very happy that you are still trying. I believe we van crack it sometimes :)

rneher:
It actually runs fairly ok on my machine. Could one make it rerender only when one leaves field through enter or tab or similar? Or wait 2s before rerendering. It doesn't update when dragging the mitigation graph. 

In general, I am big fan of autorun!

rcbevans:
Haven't taken a look at your implementation, but I would suggest taking a look at lodash's [debounce](https://lodash.com/docs#debounce) functionality with perhaps a 2s delay, reset on each input form value change.

ivan-aksamentov:
@3b3ziz Hi Ahmad, in the end I've merged an alternative solution in #170. But thanks for your attempt and investigation! If there is something else we need to carry from your branch, feel free to open another PR.

PR #91, opened 2020-03-21
Fix missing key

arnaudNYC:
## Description

There was a warning in the console about a missing `key` in the DropdownItem component


ivan-aksamentov:
@arnaudNYC Thanks! I hated that warning but never had time to fix :)

PR #94, opened 2020-03-22
Allow for browser sourcemaps when in dev mode

btoo:
This PR addresses [this comment](https://github.com/neherlab/covid19_scenarios/issues/72#issuecomment-602071478) and allows for the proper generation of sourcemaps for an improved debugger experience in `development` mode (it's actually `!= production` because I thought this would be more future proof), which I found really useful for myself.

### DIFF
--- a/config/webpack/webpack.client.babel.ts
+++ b/config/webpack/webpack.client.babel.ts
@@ -354,7 +354,7 @@ export default {
       // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
       // @ts-ignore: BUG: This parameter is missing in @types/webpack declarations
       fileContext: 'web/content/sourcemaps',
-      noSources: true,
+      noSources: production,
     }),
 
     analyze &&


ivan-aksamentov:
@btoo @petebacondarwin I usually use ["Add folder to workspace" feature of Chromium](https://developers.google.com/web/tools/chrome-devtools/workspaces#devtools) and for me the existing maps work okay :)

But if you folks need this change, then let's add it.

Note: we don't really have anything to hide in production, as we are open source anyways, so maybe we make it `false` always to facilitate debugging prod and staging deployments?

btoo:
@ivan-aksamentov ah i can't believe i havent looked into that feature until now. thanks for the tip!

my reasoning for keeping it `true` in production was to speed up page load times as the client would no longer need to download the extra bytes for the sourcemaps. I'm not sure if this is actually an issue though - so I'll let you make the call 👍

petebacondarwin:
I am happy with the "adding files from the file-system approach" if that works. I'll try later. No need to add the source content to the source-maps in production as this will make them much larger to distribute.

PR #95, opened 2020-03-22
[housekeeping] Fix list child key props in DeterministicLinePlot

btoo:
This PR simply addresses the `Warning: Each child in a list should have a unique "key" prop.` that are caused by the `<Line />` and `<Scatter />` lists in `DeterministicLinePlot`


ivan-aksamentov:
@btoo Thanks!

PR #96, opened 2020-03-22
[housekeeping] Fix LineProps['legendType'] type-safety with <Line />

btoo:
This PR simply fixes the type error created by the `interface LineProps`' `legendType` conflicting with `rechart`'s `LineProps`' expected `legendType`s, which are comprised more specifically of string literals

Screen Shot 2020-03-21 at 8 34 31 PM

### DIFF
--- a/src/components/Main/Results/DeterministicLinePlot.tsx
+++ b/src/components/Main/Results/DeterministicLinePlot.tsx
@@ -32,7 +32,7 @@ interface LineProps {
   key: string,
   name: string,
   color: string
-  legendType?: string
+  legendType?: import('recharts').LineProps['legendType']
 }
 
 function xTickFormatter(tick: string | number): string {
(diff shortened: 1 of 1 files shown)


src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -32,7 +32,7 @@ interface LineProps {
   key: string,
   name: string,
   color: string
-  legendType?: string
+  legendType?: import('recharts').LineProps['legendType']
```
ivan-aksamentov: I am surprised this works at all. Technically, dynamic `import()` returns a promise, if I remember correctly.
Could we do a proper static `import` in the file header instead?
P.S. I don't remember if we have `type-only` import supported in our config, that would be an even better option.

ivan-aksamentov:
Thanks @btoo ,
One of my goals is to fix all the type- and linting errors at some point and to re-enable linting on build time. We've disabled it early during development so that it would be easier for scientists to iterate. So I'd appreciate if you can catch some more of these bugs :)
Added https://github.com/neherlab/covid19_scenarios/issues/101

src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import ReactResizeDetector from 'react-resize-detector'
 import { CartesianGrid, Legend, Line, ComposedChart, Scatter, Tooltip, TooltipPayload, XAxis, YAxis } from 'recharts'
+import type { LineProps as RechartsLineProps } from 'recharts'
```
ivan-aksamentov: Thanks! :)

PR #99, opened 2020-03-22
Fix crash when inputing number below 2 on ContainmentGraph

dannysepler:
## Description

If you go onto the "Transmission rel to baseline" graph, and type an "invalid" number (ie below 2), the whole page crashes. This stops the page from crashing.

If you enter a number too high (10,000), the page also crashes. This throws a warning in that scenario once the numbers get higher than 100, but doesn't stop user behavior. (Would be nice!)

I probably won't get to it into this diff, but if you bring the number to 1 and then back to 10, the y values of the graph are lost seemingly until you refresh. I don't really know why this is.

## Related issues

None

Just that one graph, as far as I know!

## Testing

Manual.

## Deploy Notes

N/A

### DIFF
--- a/src/components/Form/FormSpinBox.tsx
+++ b/src/components/Form/FormSpinBox.tsx
@@ -35,6 +35,16 @@ export function FormSpinBox<T>({
   const showError = errorMessage && isTouched
   const borderDanger = showError ? 'border-danger' : ''
 
+  function validate(value) {
+    let error;
+    if (min && value < min) {
+      error = `The input cannot be less than ${min}`
+    } else if (max && value > max) {
+      error = `The input cannot be greater than ${max}` 
+    }
+    return error;
+  }
+
   return (
     <Field>
       {() => (
@@ -53,6 +63,7 @@ export function FormSpinBox<T>({
                 min={min}
                 max={max}
                 pattern={pattern}
+                validate={validate}
               />
               {showError ? <div className="text-danger">{errorMessage}</div> : null}
             </Col>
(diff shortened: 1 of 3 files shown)


src/components/Form/FormSpinBox.tsx
```diff
@@ -30,11 +30,20 @@ export function FormSpinBox<T>({
   errors,
   touched,
 }: FormSpinBoxProps<T>) {
-  const isTouched = _.get(touched, identifier)
```
dannysepler: Not really a fan of isTouched here, since it only throws errors after your mouse clicks out of the input. (To me it's better to get the error realtime.) I'm not a front end dev though, is there a reason why we may want this behavior?

ivan-aksamentov: @dannysepler Hi Danny,  thanks so much for the contribution.

I am not a fan of `isTouched` either, but it is how author of Formik recommends to do and there was a huge figth sometimes ago on Twitter and it was established that it is the best practice. Not that our app follows any UX best practices. That's probably the only one :)

But more importantly, in our particular app, removing it makes the typing in the field quite laggy, as we now run validation on every keystroke, and, thus, dispatch `useReducer` actions. Which is a very dumb thing to do in the validation in the first place. Hope we can find a way to fix it soon. See more about it here: https://github.com/neherlab/covid19_scenarios/issues/17

Does your fix still works with `isTouched`? If yes, could you please bring it back for now?

dannysepler: ha, it'll break my heart but can do!

ivan-aksamentov: Yes, a part of me died too ;)

Issue #101, opened 2020-03-22
Fix types and linting errors


ivan-aksamentov:
Although type checking and linting was implemented in the app from the beginning, we disabled it early so that scientists could iterate faster without being bothered too much with the latest web dev novelties.

We need to apply fixes **incrementally**, a few files a day. Preferably as a part of the routine PRs.

At some point we can re-enable linting on build-time, but currently you can use these commands to run `eslint` `stylelint` `tsc` and `tslint` all together in a separate terminal instance:

```
yarn lint
yarn lint:fix
yarn lint:watch
yarn lint:fix:watch
```

Or `yarn <tool_name>:[fix|watch|fix:watch]` for a specific tool.

Current output:
```
Eslint
512 warnings found.
400 warnings potentially fixable with the `--fix` option.

tsc:
Found 85 errors.

```

Note that my linting tastes are somewhat peculiar and controversial: there is a whole lot of eslint and tslint plugins and tsc is configured very strictly. I like it this way, but most developers find it annoying. 

Feel free to post your thoughts. We can relax the default config, but I'd like to keep my hard-mode one at leas as an option.

ruisaraiva19:
@ivan-aksamentov before starting to re-enable linting on build-time/development-time, we may want to migrate from TSLint to ESLint ([TSLint is deprecated](https://github.com/palantir/tslint/issues/4534)).

Regarding linting tastes, I think we can use a more standard config without much nitpicking to be easy for everyone to contribute to the project.

Integrating these tools with the IDE is one easy way to auto-fix issues and improve the developer experience as much as possible.

ivan-aksamentov:
@ruisaraiva19 The migration is already there. I mean I am using `@typescript-eslint/*` family of packages in `eslint`. 

https://github.com/neherlab/covid19_scenarios/blob/4df9bd6905e3892b97e8df295594609f025bc141/.eslintrc.js#L24-L26

But I still keep `tslint` around because `@typescript-eslint/*`  don't yet cover some of the checks that require type information:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md

ivan-aksamentov:
When relaxing the configs make sure you understand that this is a somewhat mission critical project and in the realities of an unrolled pandemic, specialists may in one way or another rely on this tool's predictions. It would be nice for it to be of a slightly higher quality than a typical web app.

[...]

PR #84, opened 2020-03-21
Auto Scroll To View Results on Mobile

bharath6365:
## Description
1) Adds smooth scroll functionality once the Result button is clicked. 
2) Add disabled cursor for all disabled buttons (Disabled buttons had cursor:pointer before)

## Related issues
       https://github.com/neherlab/covid19_scenarios/issues/71 - Point number 4

## Impacted Areas in the application
1) Result button click.

PR #103, opened 2020-03-22
Scroll ResultsCard into view when "Run" is tapped

btoo:
This PR addresses the "Tapping 'Run' button gives no clue that there is more content appeared below" task in https://github.com/neherlab/covid19_scenarios/issues/71. Specifically, it uses a `div` wrapper around `ResultsCard` (I would've preferred to use the `Col` that already wrapped it, but getting a `ref` out of reactstrap components proved too difficult) to [`scrollIntoView(alignToTop)`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView).

It also:
- ensures that the auto-scrolling is only triggered when on a mobile layout wherein the form layout is constrained to a single column
- attempts to use [`scroll-behavior: smooth`](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior) (even though it isn't supported in Safari or IE, but it's better than nothing imo), and resets `scroll-behavior` back to whatever it was after this transition
- adds a `src/helpers/hooks.ts` as a location for storing some custom hooks which have been used for this feature. all are welcome to reuse them if they'd so please

### DIFF
--- a/src/components/Main/Main.tsx
+++ b/src/components/Main/Main.tsx
@@ -32,6 +32,7 @@ import { ScenarioCard } from './Scenario/ScenarioCard'
 import { updateSeverityTable } from './Scenario/severityTableUpdate'
 
 import './Main.scss'
+import { useScrollIntoView } from '../../helpers/hooks'
 
 export function severityTableIsValid(severity: SeverityTableRow[]) {
   return !severity.some(row => _.values(row?.errors).some(x => x !== undefined))
@@ -112,7 +113,20 @@ function Main() {
           onSubmit={handleSubmit}
           validate={setScenarioToCustom}
         >
-          {({ errors, touched, isValid }) => {
+          {({ errors, touched, isValid, isSubmitting }) => {
+            /**
+             * viewport width - we only want to scroll the ResultsCard into view if viewing on mobile devices, where the layout is only a single column
+             * @see {@link https://stackoverflow.com/a/8876069/3942699}
+             */
+            const vw = Math.max(document.documentElement.clientWidth, window.innerWidth || 0)
+            /**
+             * 992 is the width at which the layout collapses into a single column
+             * @see {@tutorial https://getbootstrap.com/docs/4.0/layout/overview/#responsive-breakpoints}
+             * 
+             * only `scrollIntoView` when `isSubmitting` goes from `true` -> `false`
+             */
+            const refOfElementToScrollIntoView = useScrollIntoView<HTMLDivElement>(!isSubmitting && (vw < 992))
+
             const canRun = isValid && severityTableIsValid(severity)
 
             return (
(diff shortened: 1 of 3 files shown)


ivan-aksamentov:
Hmm, I just recently merged https://github.com/neherlab/covid19_scenarios/pull/84

Now, two solutions look quite different. What are the differences and which one is better?
Also, I like hooks, of course :) but previous solution works without hooks. So now I am confused.

I'll stop by again to re-review both solutions with a fresh head. 

P.S. Crazy mod: Github shows this PR as mergeable. What will happen if I merge this one too? :)

btoo:
Well I'm of course going to be biased 😅 but the case I'm going to make in favor of this PR is:
- https://github.com/neherlab/covid19_scenarios/pull/84 will only `scrollToViewGraphs` when `.run-button` `onClick`s, but that doesn't necessarily mean that the charts have finished rendering due to the expensive task of calculating, rendering, and (in the case of code-splitting) even loading the library+data. Without the charts rendered, there will be no content at the bottom of the viewport for `scrollToViewGraphs` to scroll to. https://github.com/neherlab/covid19_scenarios/pull/84 attempts to address this with a hardcoded guess of [`800`ms](https://github.com/neherlab/covid19_scenarios/pull/84/files#diff-1fb701f214095d223942f688e6ed4497R36), but this PR addresses this risk instead by only auto-scrolling _after_ [`isSubmitting` goes from `true` to `false`](https://github.com/neherlab/covid19_scenarios/blob/master/src/components/Main/Main.tsx#L102), which itself should only happen [synchronously _after_ `result` is set](https://github.com/neherlab/covid19_scenarios/blob/master/src/components/Main/Main.tsx#L100), which I believe is more future-proof.
  - This point would also answer your question about what would happen if you merged this PR on top of https://github.com/neherlab/covid19_scenarios/pull/84: the app would attempt to scroll to the bottom immediately upon tapping "Run", and then when the results actually get rendered, the app will attempt to scroll to the bottom again.
  - Also to answer your question about the necessity of hooks, the reason I believe hooks (specifically `useEffect`) is necessary here is because we need to react to state changes. In this context, we need to react to [`isSubmitting`](https://jaredpalmer.com/formik/docs/api/formik#issubmitting-boolean)
- https://github.com/neherlab/covid19_scenarios/pull/84 uses another hardcoded guess [in the form of scrolling the viewport down a fixed `550`px](https://github.com/neherlab/covid19_scenarios/pull/84/files#diff-1fb701f214095d223942f688e6ed4497R32), but `ResultsCard` could be taller or shorter than that. PR will reliably bring the top of the `ResultsCard` to align with the top of the viewport
- https://github.com/neherlab/covid19_scenarios/pull/84 did decide to use [`behavior: 'smooth'`](https://github.com/neherlab/covid19_scenarios/pull/84/files#diff-1fb701f214095d223942f688e6ed4497R34) instead of [`scroll-behavior: smooth`](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior), which I thought was actually a better decision (even though they both use the same underlying spec, so they are both incompatible with Safari + IE), so I decided to integrate it in this PR 👍

ivan-aksamentov:
@btoo Okay, that was a fantastic explanation. I am convinced.
Could you please help me to replace the old solution with yours? We need to ensure that they don't collide. Feel free to take whatever else pieces you may need (like that `behavior: 'smooth'`).

btoo:
ok @ivan-aksamentov i've updated the branch by simply removing the redundant usage of `scrollToViewGraphs` (which i found to also fix 2 type error messages for https://github.com/neherlab/covid19_scenarios/issues/101)

ivan-aksamentov:
@btoo I am enjoying your PRs very much. Thank you!

PR #104, opened 2020-03-22
Add favicon, increase size for range selector.

bharath6365:
Description
Add favicon for the application. (Uses the same as that of the parent domain). After this change favicons won't work locally but I guess that's no big deal.
Increase the size of input type number thereby giving users more room to click. I do it only for the first fold and not for Severity Assumptions as there's no room if the input exceeds a certain value. Hence the CSS scoping.

number-size-increase

Impacted Areas in the application
Favicon
Forms

### DIFF
--- a/src/styles/global.scss
+++ b/src/styles/global.scss
@@ -86,6 +86,17 @@ p {
 input[type='number']::-webkit-inner-spin-button,
 input[type='number']::-webkit-outer-spin-button {
   opacity: 1;
+  
+}
+
+.form-group {
+  input[type='number']::-webkit-inner-spin-button,
+  input[type='number']::-webkit-outer-spin-button {
+    position: relative;
+    right: -28px;
+    width: 40px;
+    height: 40px;
+  }
 }
 
 .noselect {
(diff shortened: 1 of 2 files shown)


src/styles/global.scss
```diff
@@ -86,6 +86,17 @@ p {
 input[type='number']::-webkit-inner-spin-button,
 input[type='number']::-webkit-outer-spin-button {
   opacity: 1;
+  
+}
+
+.form-group {
+  input[type='number']::-webkit-inner-spin-button,
```
ivan-aksamentov: This makes spin buttons disappear on Chrome. Could you please remove this?

ivan-aksamentov:
@bharath6365 favicons are welcome.
The input style changes cause arrow buttons to disappear. This is not good.

bharath6365:
@ivan-aksamentov  Did you get a chance to check?

ivan-aksamentov:
@bharath6365 yep, see comments above.

ivan-aksamentov:
@bharath6365 It's been a while without updates. I will close this for now for logistics purposes, but feel free to open another PR with the changes above. Thanks!

Issue #38, opened 2020-03-18
Support developing on Windows


ivan-aksamentov:
Currently we only support Unix-based platforms and the dev configuration contains some Unix-isms that Windows cannot digest.

We want to support running the app on windows as well, to reach more contributors.

See also:
 - https://github.com/neherlab/covid19_scenarios/issues/18#issuecomment-600847215

jamesgecko:
Carrying over the conversation from that ticket, this line is indeed problematic.
https://github.com/neherlab/covid19_scenarios/blob/00a3b2f3337abc05a78ad339898d30b099316cce/config/nodemon/dev.json#L32

It's an easy fix; that last bit of the command should look like `yarn dev:nowatch ; true"`, with a space before the semicolon.

The next thing making the Windows build fail is that `package.json` is referencing bash scripts. There are two versions of each script in `node_modules/.bin`. For example, `webpack-dev-server` and `webpack-dev-server.cmd`. The `.cmd` ones are scripts for Windows, the ones without an extension are for bash. `yarn webpack-dev-server` knows which script to run. But in `package.json`,  commands like `dev:nowatch` directly reference specific bash scripts in that directory.

My first thought was that there might be a way to get get yarn to tell us which script we need. But while `yarn bin webpack-dev-server` provides a path the to the script referenced, it doesn't take platform into account.

AmadeusW:
If someone wants to develop on Windows right now, before this issue is resolved: consider enabling [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/about) for development.

rcbevans:
I'll look into this now

ivan-aksamentov:
@rcbevans Hi Rich, thanks!
If you'll be looking into this, let me know if you have any other suggestions along the way, regarding the build, dev experience or otherwise.

ivnnv:
Hi,
I have been successfully able to run the project on **Windows 10** (if you are not yet on Windows10, you should), and using [Windows Subsystem for Linux]( https://docs.microsoft.com/en-us/windows/wsl/about) + Ubuntu bash on VSCode.

![wls vscode](https://user-images.githubusercontent.com/23552631/77234397-66b76d00-6bae-11ea-8a31-f4629ddba19f.png)

[...]

PR #114, opened 2020-03-22
#38 support windows development

eloisetaylor5693:
#38 Added support for Windows development.

I didn't test every npm script.

Side note:  I think the script should be called `start` instead of `dev`.  eg `yarn start` or `yarn serve:dev`


config/webpack/webpack.client.babel.ts
```diff
@@ -145,7 +145,7 @@ export default {
     compress: true,
     disableHostCheck: true,
     historyApiFallback: true,
-    host: '0.0.0.0',
+    host: '127.0.0.1',
```
ivan-aksamentov: It will break for people who work on remote machines. Why are these necessary?
It works for me with `0.0.0.0` on Windows 7 VM.
If it is really necessary, please make it configurable through an environment variable and add to `.env.example` file

package.json
```diff
@@ -5,10 +5,12 @@
...
   "license": "MIT",
   "scripts": {
+    "predev": "node -e 'console.clear()' && yarn install && yarn dev:clean",
     "dev": "nodemon --config config/nodemon/dev.json",
-    "dev:nowatch": "cross-env NODE_ENV=development BABEL_ENV=development babel-node node_modules/.bin/webpack-dev-server --config config/webpack/webpack.client.babel.ts",
+    "dev:nowatch": "cross-env NODE_ENV=development BABEL_ENV=development webpack-dev-server --config config/webpack/webpack.client.babel.ts",
     "dev:clean": "rimraf '../.build/development/{*,.*}' 'node_modules/.cache'",
-    "prod": "cross-env NODE_ENV=production BABEL_ENV=production babel-node node_modules/.bin/webpack --config config/webpack/webpack.client.babel.ts",
+    "preprod": "node -e 'console.clear()' && yarn install && yarn prod:clean",
+    "prod": "cross-env NODE_ENV=production BABEL_ENV=production webpack --config config/webpack/webpack.client.babel.ts",
```
ivan-aksamentov: Using pre- hooks only makes things even more confusing in my already super-confusing setup. =)
Could we avoid that? As Microsoft folks explained in the issue thread, it is sufficient to just add spaces around semicolons `;`.
In fact I was also working on it in parallel and it indeed works.

ivan-aksamentov:
Hi @eloisetaylor5693 

I've been working on this slowly in parallel too and I decided to crowdtest it here:
https://github.com/neherlab/covid19_scenarios/issues/38#issuecomment-602241783

and in the end merged my changes, which I found more comprehensive and less breaking.
Here is a summary:
https://github.com/neherlab/covid19_scenarios/compare/28b1468..de97b3b

I will close this for now. But if you still need to adjust something, feel free to jump in to #38 or in the chat or to make another PR

config/webpack/webpack.client.babel.ts
eloisetaylor5693: On both my windows machines, 0.0.0.0 didn't work.  127.0.0.1 is the standard localhost IP address.

ivan-aksamentov: @eloisetaylor5693 Could you please patch and verify it for us?
Add and `.env` variable and set it to `127.0.0.1` by default. Refer to this variable in webpack config.  Those who need remote will change the var to `0.0.0.0`,
Thanks for your investigation!

PR #123, opened 2020-03-22
Adding D3 Types

jsmith:
## Description
Adding `@types/d3` and fixing the associated type errors. There are minimal code changes except for a few locations which will I will point out below. Everything else is just type changes (and a few linter changes)!

## Related issues
- Contributes to #101

Most changes were concentrated in the ContainmentGraph but were primarily type changes.

## Testing
No tests were added as no functionality was added/modified.


src/components/Main/Containment/ContainmentGraph.tsx
```diff
@@ -21,9 +21,7 @@ function draw({ data, width, height, onDataChange, d3ref }: DrawParams) {
   if (d3ref.current === null) {
     return
   }
-  const newData = data.map(d => {
-    return { t: d.t, y: d.y }
-  }) // copy
+  const newData = data.map((d): TimePoint => ({ t: d.t, y: d.y })) // copy
```
jsmith: This function is now a one-liner instead of a full function with a return statement.

src/components/Main/Containment/ContainmentGraph.tsx
```diff
@@ -162,57 +160,71 @@ function draw({ data, width, height, onDataChange, d3ref }: DrawParams) {
...
+          .selectAll('tspan')
+          .data(value.split(/\n/))
+          .join('tspan')
+          .attr('x', 0)
+          .attr('y', (d, i) => `${i * 1.1}em`)
+          .style('font-weight', (_, i) => (i ? null : 'bold'))
+          .text(d => d),
+      )
+
+    const node = text.node()
```
jsmith: Here I have to handle the situation where `text.node()` returns `null`. I decided to throw an error as this would certainly be a bug. Previously, `text.node()` was assumed to always return an object.

PR #126, opened 2020-03-22
fix(form): make form help button align with input label

victor-cordova:
## Description

Hello, I noticed that the ? icon was slightly misaligned with the form label so I decided to take a look.

The header of the different sections and the form fields

## Testing

I compared how it looks in mobile and desktop viewports and it looks good

### Before

before

### After

after


ivan-aksamentov:
Hi Victor @victor-cordova, thanks, this looks so much better now!
Let me know if you have further suggestions regarding styling or anything else.

PR #129, opened 2020-03-22
fix syntax error in SeverityCard

btoo:


ivan-aksamentov:
@btoo  Whoops, faulty merge! :)
But I fixed on master already

PR #135, opened 2020-03-23
fix AgeBarChart type errors

btoo:
## Description

`react-i18next`'s translation function (`useTranslation().t`) returns `TFunctionResult` (i.e. `string | object | Array<string | object> | undefined | null`), but `rechart`'s components only accept `string | number | undefined` for their `name`s, `label`s, .etc, so extend `TFunction` with a signature that returns _just_ `string | number | undefined`

## Related issues

https://github.com/neherlab/covid19_scenarios/issues/101

### DIFF
--- a/src/components/Main/Results/AgeBarChart.tsx
+++ b/src/components/Main/Results/AgeBarChart.tsx
@@ -24,7 +24,16 @@ export function AgeBarChart({ data, rates }: SimProps) {
     return null
   }
 
-  const { t } = useTranslation()
+  const { t: unsafeT } = useTranslation()
+  const t = (...args: Parameters<typeof unsafeT>) => {
+    const translation = unsafeT(...args)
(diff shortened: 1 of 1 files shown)


src/components/Main/Results/AgeBarChart.tsx
```diff
@@ -11,6 +11,7 @@ import { SeverityTableRow } from '../Scenario/SeverityTable'
 import { colors } from './DeterministicLinePlot'
```
ivan-aksamentov: It is basically a complicated way to cast return value to a more narrow type.
But what if  someone on the other end of the translation pipe goes crazy and `t()` will actually return an object or an array? Everything will blow up.

I think this should be a runtime check with a throw. Not elegant, but safe.
Feel free to either one-off inline or to wrap into a function (`safeT()`? `tString()`? idk) with the correct return type, both compile-time and run-time.

Issue #134, opened 2020-03-23
In Safari help text does not appear on clicking the help buttons


ivan-aksamentov:
# 🐛 Bug Report

In Safari help text does not appear on clicking the help buttons

This was reported by @Johan-A-M in chat:
https://spectrum.chat/covid19-scenarios/general/questions-discussions~8d49f461-a890-4beb-84f7-2d6ed0ae503a?m=MTU4NDkwODk0NTM5MQ==

## 🤔 Expected Behavior

Clicking on help button should bring up the popup with explanation text

## 😯 Current Behavior

Apparently, nothing happens on click when browsing in Safari on Mac

## 💁 Possible Solution

Unknown

## 🔦 Context

Not specified

## 💻 Code Sample

Not specified

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Browser          | Safari, unspecified version
| Operating System | Mac, , unspecified version
| NPM/Node/Yarn    |

Related:
https://github.com/neherlab/covid19_scenarios/issues/20

gj262:
Sounds like: https://github.com/reactstrap/reactstrap/issues/1428

ivan-aksamentov:
@gj262 Good catch! Do you have a possibility to test on Safari?

rebmullin:
hi @ivan-aksamentov, I just created a patch for this. Please let me know if anything is needed. thanks :)

gj262:
@ivan-aksamentov  I do. It looks like many people are on this bug! I'll see how those PRs go.

rebmullin:
I didn't realize so many people were working on this one! i can close mine if needed too:)

gj262:
Hi @rebmullin I tried out your PR but it does not work for my safari. I'll give #138 a go.

ivan-aksamentov:
@rebmullin @gj262 That was a fix response to user's bug report :)

gj262:
I took a look at: http://reactstrap.github.io/components/popovers/#Popovers-Trigger in safari. The 'legacy' and 'click' methods work and perhaps we can simply use one of those for the Popover. I'll give that a try.

gj262:
Ok. The following works with chrome and safari. 

- Use an UncontrolledPopover rather than a Popover.
- Use trigger="legacy"
- Keep the onClick prop on Button to avoid triggering dropdown toggle for labels that have them. i.e. Results.

[...]

PR #138, opened 2020-03-23
redo FormHelpButton trigger

btoo:
## Description

Due to https://github.com/reactstrap/reactstrap/issues/1428 I had to replace the dependency on `onFocus` with a handler that would display the tooltip when the button is clicked, and then hide it when the user clicks outside of the tooltip

(This PR also performs `prettier --write` on `src/helpers/hooks.ts`)

## Related issues

https://github.com/neherlab/covid19_scenarios/issues/134

### DIFF
--- a/src/components/Form/FormHelpButton.tsx
+++ b/src/components/Form/FormHelpButton.tsx
@@ -1,10 +1,11 @@
-import React, { useState } from 'react'
+import React, { useState, useRef } from 'react'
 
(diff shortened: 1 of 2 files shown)


ivan-aksamentov:
I accepted #139 for now. Let me know if you want to ammend.

ivan-aksamentov:
@btoo That fix did not work. Plus there are some additional concerns. Could you jump in to https://github.com/neherlab/covid19_scenarios/issues/134 with your ideas?

PR #152, opened 2020-03-23
amend german translation

FireLizard:
Contributes to https://github.com/neherlab/covid19_scenarios/issues/56


src/locales/de/translation.json
```diff
@@ -57,7 +59,9 @@
   "Mitigation": "Linderung",
```
ivan-aksamentov: Could you please address Richard's suggestion:
> I think a better translation of `Mitigation` in this context might be simply `Maßnahmen`.
https://github.com/neherlab/covid19_scenarios/pull/152#pullrequestreview-379309257
?

FireLizard: Sure :+1: 
Sorry for late response

PR #153, opened 2020-03-23
fix key for death count in plot set-up

rneher:
## Description
fix problem with wrong key in plot-data set up. That said, or nomenclature regarding `dead` `death` and `deaths` is for from ideal. 

Only affects the data aggregation for the plot.


rcbevans:
For nomenclature I'd have expected to see something like 'Mortality', rather than Death

Issue #120, opened 2020-03-22
Flexible time range for population presets


rneher:
# 🙋 Feature Request
The top left panel contains population parameters and the simulation time range. The latter partly belongs to the population (the start of the outbreak) and is partly depends on the question the user is asking.  tMin in particular should be tied to the initial case count. 

## 🤔 Expected Behavior
The initial condition should be flexible both with regard to case count and time. Wuhan for example needs 100 cases in Dec 19. 

## 😯 Current Behavior
Hard coded to March 1st

## 💁 Possible Solution
Split the `populationData.tsv` into a set of fixed population parameters (size, hospital beds, ageDistribution key) and an `inititalCondition.json`.  

We could either parse `initialCondition` could be parsed into the `src/assets/data/scenarios/population.json` or make 

https://github.com/neherlab/covid19_scenarios/blob/master/src/assets/data/scenarios/simulation.ts

population specific in analogy to 

https://github.com/neherlab/covid19_scenarios/blob/master/src/assets/data/scenarios/population.ts

We could tie population and simulation presets to the same default scenario key. This would require only few changes in the rest of the code base.

nnoll:
An additional thought on this overhaul is that it would be nice to have the % confirmed cases also toggle-able by scenarios. This could be estimated from our case/death data just as tMin and the number of initial cases.

PR #155, opened 2020-03-23
make simulation time range depend on population presets

rneher:
Outbreaks in different population started at different times and it would be useful to have flexible presets reflecting this. This would also allow for better comparisons and case counts. 

## Description
ingests as `simulation.json` in `src/assets/data/scenarios/simulation.ts` and creates an object of simulationData presets associated with each region. Previously, this was a global hard-coded preset for all. This doesn't play terribly well with the architecture of our scenario presets. 

## Related issues
Contributes to https://github.com/neherlab/covid19_scenarios/issues/120

Making the simulationData dependent on the population requires multiple changes in reducers, state, and data.  `getSimulationData()` now takes the name of the population as an argument. The initialization of the containmentGraph needs to know about the population.

## Testing
Currently, the form `isValid` evaluates to false when the population preset doesn't have `caseCounts`. Otherwise it seems to work, but needs cleaning up.


ivan-aksamentov:
@rneher Looks like it needs a good rebase-reshuffle... We are moving fast :)

ivan-aksamentov:
Will be superseded by another PR. Closing after discussion with @rneher

PR #156, opened 2020-03-23
#101/Fix type conflict caused by global exports of Chai, a Cypress dependency.

abrie:
## Description

This fix addresses a specific class of type errors in the unit tests. Both Jest and Chai (a Cypress dependency) declare a global "Assertion" type. Under certain conditions the Chai version takes precedence, resulting in the type error. For example: 

`error TS2339: Property 'toBeObject' does not exist on type 'Assertion'.`

This fix works by excluding the cypress/ folder from `tsconfig.json`, thereby keeping the unnecessary Chai definitions out of the global namespace.

## Related issues

#101 

Unit tests.

## Testing

`yarn test:lint`

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -26,6 +26,26 @@
     "sourceMap": true,
     "strict": true
   },
-  "exclude": [".build", ".cache", ".ignore", ".reports", "3rdparty", "node_modules", "static", "tsconfig.json"],
-  "include": ["./.js", "./.jsx", "./.ts", "./.tsx", "./**/*.js", "./**/*.json", "./**/*.jsx", "./**/*.ts", "./**/*.tsx"]
+  "include": [
+    "./.js",
+    "./.jsx",
+    "./.ts",
+    "./.tsx",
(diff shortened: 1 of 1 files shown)


gj262:
Good fix. I was looking at this also and came up with the same approach just now :)

It also clears up a lot of the test file errors in the `yarn lint` -> `yarn tsc` targets.

abrie:
@gj262 Thanks – I was pretty frustrated while solving the issue. The linter did not reliably report the result unless a source file was modified, and the slow tsc/linter step was no help. Does that mirror your experience?

gj262:
> @gj262 Thanks – I was pretty frustrated while solving the issue. The linter did not reliably report the result unless a source file was modified, and the slow tsc/linter step was no help. Does that mirror your experience?

Pretty much. I also found that restarting the typescript environment in my IDE (vscode) was necessary to reliably detect the errors.

ivan-aksamentov:
Thanks for the investigation!

PR #161, opened 2020-03-23
Model JS -> TS Conversion

jsmith:
## Description
The goal of this PR is to convert the model.js file from JavaScript to TypeScript. This is the only .js file in the repo and it seems like a good idea to convert it to TypeScript since it is some of the core code.

## Related issues
Contributes to #101

Changes mostly affect the model.js file with a few small changes in other locations.

## Testing
Some of these functions are a bit hard to test to I decided to take a few JSON snapshots, put them in the assets folder and load them during unit tests. These snapshots were created using the default parameters. I created these snapshots before making the changes so I am fairly confident that the TS file has the same functionality as the old JS file.

One more thing, git sadly thinks I deleted model.js and created model.ts so I created [a diff of model.js and model.ts](https://www.diffchecker.com/dmbr6HZR) to make my changes easier to review. I'll also point out where I had to change the code to make it easier to type check.

Let me know your thoughts about my changes and my additional tests!


src/algorithms/model.ts
```diff
@@ -0,0 +1,376 @@
...
+import { SeverityTableRow } from '../components/Main/Scenario/SeverityTable'
+import { ModelParams, SimulationTimePoint } from '../algorithms/types/Result.types'
+
+const msPerDay = 1000 * 60 * 60 * 24
+
+const monthToDay = (m: number) => {
+  return m * 30 + 15
+}
+
+const jan2020 = new Date('2020-01-01').valueOf() // time in ms
```
jsmith: This was using implicit `Date` -> `number` conversion so I am now explicitly converting to a `number`.

src/algorithms/model.ts
```diff
@@ -0,0 +1,376 @@
...
+    timeDeltaDays,
+    // Dummy infectionRate function. This is set below.
+    infectionRate: () => -Infinity,
+    timeDelta: msPerDay * timeDeltaDays,
+    incubationTime: params.incubationTime,
+    populationServed: params.populationServed,
+    numberStochasticRuns: params.numberStochasticRuns,
+    hospitalBeds: params.hospitalBeds,
+    ICUBeds: params.ICUBeds,
+  }
```
jsmith: The initialization is now occurring in a single line rather than adding fields one my one.

src/algorithms/model.ts
```diff
@@ -0,0 +1,376 @@
...
+
+  // Population average rates
+  pop.recoveryRate.total = (1 - hospitalizedFrac) / params.infectiousPeriod
+  pop.hospitalizedRate.total = hospitalizedFrac / params.infectiousPeriod
+  pop.dischargeRate.total = (1 - criticalFracHospitalized) / params.lengthHospitalStay
+  pop.criticalRate.total = criticalFracHospitalized / params.lengthHospitalStay
+  pop.deathRate.total = fatalFracCritical / params.lengthICUStay
+  pop.stabilizationRate.total = (1 - fatalFracCritical) / params.lengthICUStay
+  pop.overflowDeathRate.total = (params.overflowSeverity * fatalFracCritical) / params.lengthICUStay
+  pop.isolatedFrac.total = avgIsolatedFrac
```
jsmith: Previously, `pop` was being initialized to `{ ...params }` so it had access to all of fields that `params` had. Now that the initialization of `pop` has changed, I had to grab a lot of information from `params` instead.

src/algorithms/model.ts
```diff
@@ -0,0 +1,376 @@
...
+
+// NOTE: Assumes all subfields corresponding to populations have the same set of keys
+export function evolve(pop: SimulationTimePoint, P: ModelParams, sample: (x: number) => number): SimulationTimePoint {
+  const sum = (dict: Record<string, number>): number => {
+    return Object.values(dict).reduce((a, b) => a + b, 0)
+  }
+  const fracInfected = sum(pop.infectious) / P.populationServed
+
+  const newTime = new Date(pop.time + P.timeDelta)
+  const newPop: SimulationTimePoint = {
```
jsmith: Previously, a number was being passed to a function which took in a date; however, implicit type coercion was allowed this. Now we are explicitly creating a date!

src/algorithms/model.ts
```diff
@@ -0,0 +1,376 @@
...
+    intensive: {},
+    dead: {},
+  }
+
+  const push = <Sub extends NumberRecordKeys<SimulationTimePoint>>(sub: Sub, age: string, delta: number) => {
+    // To get TS to type check this function, we need first assign newPop[sub] to Record<string, number>
+    // There is possibly a better solution
+    const record: Record<string, number> = newPop[sub]
+    record[age] = pop[sub][age] + delta
+  }
```
jsmith: This function was tricky to type check; however, other than splitting the body into two lines, nothing else changed.

rneher:
Thanks a lot. This produces the exact same output in my hands and will make further changes more robust. (some of us only started learning TS and stuff needed to move quickly, hence the mix...)

jsmith:
No problem :)

Issue #107, opened 2020-03-22
Add landing page


ivan-aksamentov:
Currently the application just dumps user right into the simulator. No explanations, no context, just a form and a "Run" button.

We would like to add a Landing page:
 - [x] spacey, lots of color, curvy SVG lines and a large SVG image of coronavirus (if available, let's discuss options/licensing/costs)
 - [x] a large button in the middle to launch a simulator (e.g. "Simulate")
 - [ ] basic intro in a few words: what, why, who. Not too much science. Not boring.
 - [ ] refer to science details with a link
 - [x] a checkbox to skip the landing page after the first visit (saved in the local storage)

tryggvigy:
I'm no designer but is this close to what you have in mind?

| Mobile | Desktop |
| -| -|
| ![Google Pixel 2 - 2](https://user-images.githubusercontent.com/2373958/77256180-77c2b580-6c64-11ea-8882-ef11e40077fa.png) | ![MacBook Pro - 1](https://user-images.githubusercontent.com/2373958/77256177-76918880-6c64-11ea-9529-507fd7574162.png)

There is plenty of space to move things around and add more content

ivan-aksamentov:
@tryggvigy Holy Molly, this is gorgeous!
I will merge the same second you submit a PA :heart_eyes:

tryggvigy:
Great! I'll start working on a PR as soon as I can.

---------

> basic intro in a few words: what, why, who. Not too much science. Not boring.
> refer to science details with a link

Any thoughts already on exact text we want in these and other sections of the landing page?

ivan-aksamentov:
@tryggvigy You can play with what the Overview section in the Readme has
https://github.com/neherlab/covid19_scenarios/blob/master/README.md#overview
It will probably need to be slightly less scientific and somewhat shorter. Feel free to rephrase and to adjust the length to mobile use-case.
We can link to About page and FAQ, but I'm afraid we may lose users there, as they will not know where is simulator now. Links and navigation have to be planned carefully too.

tryggvigy:
I got some progress on this today. https://github.com/neherlab/covid19_scenarios/pull/163

@ivan-aksamentov do you think we should have the landing page be a separate url from the simulator? It would require us to change the routing a bit. What I mean is something like:

The landing page would be https://neherlab.org/covid19/
and the simulator would live under https://neherlab.org/covid19/simulate.

[...]

PR #163, opened 2020-03-23
Landing Page

tryggvigy:
## Description
 Add a landing page that introduces the tool. The goal is to easy users into the tool and make the initial experience less jarring.

- On every visit the user is redirected to the landing page (`/` -> `/start`)
- Once on the landing page the user and the user clicks "Simulate", the redirect will never happen again for that session (until a refresh or a real browser navigation happens)
- The user can decide to check the checkbox at the bottom of the page before clicking "Simulate". This will cause the landing page redirect to be permanently disabled. In addition the checkbox will not be present on the landing page if the user chooses to directly navigate to it.
  -  image

### Responsiveness
foo

### Checklist
- [x] Compress image  with https://squoosh.app/
- [x] Responsive: Breakpoints for Mobile and Tables
- [x] Checkbox to skip landing page on subsequent visits.
- [x] Less scientific introductory text section built from the [Overview section](https://github.com/neherlab/covid19_scenarios/blob/master/README.md#overview) of the readme
- [ ] Pixel push
  - See if I can do the svg curve in a smarted way. I'm not great with svg
  - fix the gradient color. It looks darker and not as good, but this is the linear gradient code Figma says corresponds to my design
  - lots more

## Related issues
- Fixes https://github.com/neherlab/covid19_scenarios/issues/107

## Testing

TODO


rsharvill:
@tryggvigy I created a user guide in the README.md that I think would be helpful to include on the landing page. Let me know if you would like me to add this once this pull request is approved

src/components/LandingPage/LandingPage.scss
```diff
@@ -0,0 +1,142 @@
...
+    position: unset;
+    display: block;
+    width: 100%;
+    margin-bottom: 20px;
+  }
+
+  .landing-page__content-section {
+    padding: 20px 40px;
+  }
+}
```
tryggvigy: I'm not happy with the state of this CSS. I might spend some more time to get it into better shape. e.g. not use grid, consolidate breakpoints, contain weirdness due to the svg wave better, less magic numbers, use more variables, etc.

tryggvigy:
@ivan-aksamentov This is now functionally ready. 

Would love a review before writing the tests in case something big needs to change.

tryggvigy:
Yeah sounds good 👍

PR #168, opened 2020-03-24
chore: spanish translation

emmurillo:
## Description

Provides Spanish translation

## Related issues

- Contributes to https://github.com/neherlab/covid19_scenarios/issues/56


ivan-aksamentov:
Muchas gracias señor Sánchez!

Just to clarify, is this Spanish like in Spain, or some Latin-American flavor (sorry if the question is dumb, I don't know any better :) )
Would there be much difference? (that is should we introduce language flavo(u)rs?)

emmurillo:
Con gusto!

I tried my best to keep it neutral so people from Spain or Latin American can understand it.. I don't think flavors are needed for Spanish at least 🙂

PR #171, opened 2020-03-24
performance: only render results if they shallowly change

gj262:
## Description

When the scenario form is interacted with the entire wrapped <Formik> is re-rendered. There is a sluggishness that is especially noticeable when the run results are displayed. This change causes the ResultsCard to only re-render when a prop shallowly changes via React.memo. 

Note: The only change in this diff that matters is the `React.memo` export. The rest are auto-save prettier lint fixes.

## Related issues

Related to https://github.com/neherlab/covid19_scenarios/issues/17

Form performance.

## Testing

Run. Then interact with the form. It should be reasonably performant.

### DIFF
--- a/src/components/Main/Main.tsx
+++ b/src/components/Main/Main.tsx
@@ -35,11 +35,11 @@ import './Main.scss'
 import { useScrollIntoView } from '../../helpers/hooks'
 
 export function severityTableIsValid(severity: SeverityTableRow[]) {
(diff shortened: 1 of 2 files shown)


rneher:
thanks a ton. This is big improvement. It should also make the autorun feature https://github.com/neherlab/covid19_scenarios/pull/170 much more useful.

Issue #142, opened 2020-03-23
Scrolling position is remembered incorrectly after page reload


ivan-aksamentov:
# 🐛 Bug Report

I don't have the exact steps yet. It is especially noticeable on a narrow device. Also when dev tools are open wide and the viewport is narrower and can be scrolled.

## 🤔 Expected Behavior

Remembered scrolling position is reset after page reload, preferably.
Otherwise, maybe at least reset after I scroll to the top?
Lack of memory whatsoever would also be fine. We don't need fancy.

## 😯 Current Behavior

Scroll position is being retained needlessly and unexpectedly. Also wrongly (the position is not the same as)

After scrolling and then refreshing the page (F5, Ctrl+F5) the browser scrolls to some place, not necessarily where I scrolled previously. After scrolling to the top page and refreshing it is still scrolling somewhere.

## 💁 Possible Solution

Suspected offender PRs:
https://github.com/neherlab/covid19_scenarios/pull/84
https://github.com/neherlab/covid19_scenarios/pull/103

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Browser          |
| Operating System |
| NPM/Node/Yarn    |

rneher:
this behavior goes away if the condition here

https://github.com/neherlab/covid19_scenarios/blob/master/src/components/Main/Main.tsx#L129

evaluates to false.

ivan-aksamentov:
Fixed in #179

PR #179, opened 2020-03-24
#142/Fix unexpected scroll behavior.

abrie:
## Description

This PR refactors/re-implements the conditional scrolling feature. When a simulation is run, it triggers a hook in `src/Main/Results/ResultsCard.tsx' to update button states and scroll position. When the page is reloaded, the scroll position is set according to default browser behavior. 

Unlike the previous implementation, this one makes no attempt to detect a narrow screen. Scrolling is always performed when a simulation is run (including auto-run). This might not be desirable, but can be addressed with another PR.

## Related issues

#142, #103, #84

This affects the UI, specifically the Main Page and ResultsCard.

## Testing

This must be tested manually by running simulation scenarios with different window sizes.


src/components/Main/Main.tsx
```diff
@@ -30,7 +30,6 @@ import { ScenarioCard } from './Scenario/ScenarioCard'
 import { updateSeverityTable } from './Scenario/severityTableUpdate'
 
 import './Main.scss'
-import { useScrollIntoView } from '../../helpers/hooks'
```
abrie: useScrollIntoView is deprecated by this PR, and the hook has been removed.

src/components/Main/Main.tsx
```diff
@@ -132,18 +131,6 @@ function Main() {
...
-            /**
-             * viewport width - we only want to scroll the ResultsCard into view if viewing on mobile devices, where the layout is only a single column
-             * @see {@link https://stackoverflow.com/a/8876069/3942699}
-             */
-            const vw = Math.max(document.documentElement.clientWidth, window.innerWidth || 0)
-            /**
-             * 992 is the width at which the layout collapses into a single column
-             * @see {@tutorial https://getbootstrap.com/docs/4.0/layout/overview/#responsive-breakpoints}
-             *
-             * only `scrollIntoView` when `isSubmitting` goes from `true` -> `false`
```
abrie: This implementation does account for window size, so this code is removed.

src/components/Main/Results/ResultsCard.tsx
```diff
@@ -62,9 +62,32 @@ function ResultsCardFunction({ canRun, autorunSimulation, toggleAutorun, severit
...
+      }
+    } else {
+      setHasResult(false);
+      setCanExport(false);
+    }
+  }, [result, scrollTargetRef.current])
+
   return (
+    <>
+    <span ref={scrollTargetRef}></span>
```
abrie: A small empty element serves as the target for scrolling. This is necessary because of the way React treats refs in functional components.

src/components/Main/Results/ResultsCard.tsx
```diff
@@ -62,9 +62,32 @@ function ResultsCardFunction({ canRun, autorunSimulation, toggleAutorun, severit
...
+          block: 'start',
+          inline: 'nearest'
+        })
+      }
+    } else {
+      setHasResult(false);
+      setCanExport(false);
+    }
+  }, [result, scrollTargetRef.current])
+
```
abrie: This implementation provides a central authority for scroll behavior. The hook watches for changes on the Result object. Button states and scrolling behavior are computed accordingly.

ivan-aksamentov:
@abrie Thanks for relieving us from this bug, I think all devs are very thankful :)

PR #180, opened 2020-03-24
test: Fix regression failure in algorithms/test/run.test.tsx

abrie:
## Description

The country name label "United States" has been changed to "United States of America", but the field was not updates in the unit test. This PR addresses that fragility. With this PR we simply run all the population scenarios. It measures the same functionality, but removes the need to sync data labels in the unit test.

## Related issues

The breaking change was introduced in 5594f13007f9f95b854a8a60efb0af0dc314dd42

Unit tests.

## Testing

`yarn test:lint src/algorithms/test/run.test.tsx`

### DIFF
--- a/src/algorithms/test/run.test.tsx
+++ b/src/algorithms/test/run.test.tsx
@@ -42,48 +42,10 @@ describe('run()', () => {
   })
 
   it('should work for a lot of countries', async () => {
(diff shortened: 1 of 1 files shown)


rneher:
Thank you!

PR #191, opened 2020-03-24
added kosovo's age distribution

nnoll:
## Description

Very simple addition of Kosovo's age distribution into our json.


src/assets/data/country_age_distribution.json
```diff
@@ -1066,6 +1066,17 @@
     "70-79": 2331,
     "80+": 590
   },
+  "Kosovo": {
```
rcbevans: When new data is added, is the source of the data cited anywhere?

rcbevans: Total of these numbers, 1.78M is slightly lower than a few sources, e.g. [CIA World Factbook](https://www.cia.gov/library/publications/the-world-factbook/geos/kv.html) which has the population at ~1.93M, [World Bank](https://data.worldbank.org/country/kosovo) at 1.85M

ivan-aksamentov: @rcbevans I think we gather licenses and citations for case data here:
https://github.com/neherlab/covid19_scenarios_data/blob/37525e3b55/sources.json
but not sure about population and beds data.
@nnoll Nick, could you please comment?

Issue #182, opened 2020-03-24
Modify the model to allow for non-exponential latency periods.


nnoll:
# 🙋 Feature Request

I want to generalize from an exponential distribution of latency time to a modifiable Erlang distribution that more accurately recapitulates measured distributions.

## 🔦 Context

Currently the exposed class models the latency period between exposure to an individual infected with COVID-19 and becoming ill. With a singular state, the distribution of times is exponential which is a bad fit to the data, e.g. see [here](https://annals.org/aim/fullarticle/2762808/incubation-period-coronavirus-disease-2019-covid-19-from-publicly-reported).

In short, we are both overestimating the variance and the number of quick infection cases. This makes a qualitative difference as to where the peak/apex of cases lies which is entirely relevant for hospital planning resources.

## 😯 Describe the feature

Note this is very much related to Issue #26, however the distribution is not able to be arbitrarily edited by the user. Similarly, this is related to Issue #116, however it would not require shifting to a DDE (delayed differential equation).

## 💻 Examples

N/A

## 💁 Possible Solution

We internally increase the number of exposed states to a small cascade of 2/3 states as the distribution of the sum of exponentials is an Erlang distribution. In the future, we can consider making this tunable.

nnoll:
I open this to start a discussion. If this seems good, I'll implement it.

rneher:
Yes, would be a useful addition. I don't think we would need more than 2-3 states. One could also branch this to implement superpositions of distributions of different width.

alo8:
Yes, using a cascade of states will make the simulation more realistic.
With just one exposed state we have a exponential distribution. The rate parameter for this distribution is equal to 1/T, with T being the incubation time.  
If we use a cascade of s states, we will get a Erlang distribution as pointed out by @nnoll .
The rate parameter of this Erlang distribution is s/T and the shape parameter is equal to s.
This final distribution will have a mean of T (as expected) and a variance of:
<img src="https://latex.codecogs.com/gif.latex?\frac{T^{2}}{s}" />
So if we would use an infinite number of cascades, the variance would become 0 and the Erlang distribution would become a delta distribution centered around T. This is exactly the proposal #116 (delayed differential equation). So by variation of the cascade number s, we have a way to gradually translate one proposal into the other.

But which casade numer s is the most realistic? 
[...]

PR #192, opened 2020-03-24
Adds internal states to exposed class

nnoll:
Modifies the underlying algorithm to allow for a more realistic latency distribution as per Issue #182 . This qualitatively changes the predictions so we should look at this with critical eyes. 

Note, I had to break a bit of the types that have been so nicely set up. This probably needs the attention of someone more fluent in typescript to make the linter happy.

Also includes some refactoring of the UserResult that should decouple the simulation data structure and that used for plotting.


rneher:
This looks very good. I would suggest to remove the hard coded `3` in various places. Also the terminology could use revision. `newInfectious` is only a good name for the very last exposed category. Why not call it something like `exposedFluxOut`. While you are at it, maybe change `incubationTime` to `latency`?

nnoll:
@ivan-aksamentov what needs to be done to get this merged? I'd like to update the model before this branch desyncs too much.

ivan-aksamentov:
@nnoll It needs to mergeable but also it is desireable for it to be functional. The deployed version crashes on Run. I'll see what can be done.

ivan-aksamentov:
@nnoll In this commit, where you merged master into the feature branch:
https://github.com/neherlab/covid19_scenarios/commit/711b5a4603272f8ed47791dfb0a9e2d2468c0b02
the number formatting/humanization was not merged correctly.
I added missing variables, so at least it runs now, but I am still not sure if there are other problems.
I'll merge and then we can hopefully fix if somethings remains. And tests of course.

ivan-aksamentov:
I close this dues to errors in merges. I rewound all the non-erroneous commits in #221 and will merge it.

PR #193, opened 2020-03-25
Fixing type errors in Main.tsx

jsmith:
## Description
Fixing the 3 type errors in Main.tsx. To do this, I used TypeScript type predicates to ensure the given values were a valid country and a valid region. I also fixed the `deserializeScenarioFromURL` function to correctly return a `State` object (this was causing the third error). Also let the linter format some of the files I made edits to.

## Related issues
Contributes to #101

Just `Main.tsx`. No functionality changes.

## Testing
Confirmed that the type predicates worked correctly by running a few scenarios.


src/components/Main/state/URLSerializer.ts
```diff
@@ -60,7 +61,10 @@ export function deserializeScenarioFromURL(initState: State) {
         containment: {
           ...initState.containment,
           current: obj.current.containment,
-          data: { reduction: containmentDataReduction },
+          data: {
+            reduction: containmentDataReduction,
+            numberPoints: initState.containment.data.numberPoints,
+          },
```
jsmith: @brunorzn Does this small edit look ok to you? This was causing a TS error since `numberPoints`was missing.

brunorzn: Yes, it loos good! Thanks!

jsmith:
#190 was just merged which casts `countryAgeDistributionData` as `CountryAgeDistribution` which removes the type error. This PR instead explicitly checks the presence of the country which may be beneficial as this it's possible to set the country parameter using the URL string (and possibly using other means in the future).

src/components/Main/Main.tsx
```diff
@@ -42,6 +42,14 @@ export function severityErrors(severity: SeverityTableRow[]) {
 
 const severityDefaults: SeverityTableRow[] = updateSeverityTable(severityData)
 
+const isCountry = (country: string): country is keyof typeof countryAgeDistribution => {
```
rcbevans: Interesting, never seen this syntax before.

jsmith: Yeah it's a not as well known TypeScript feature called type predicates. Super useful in some situations!

Issue #187, opened 2020-03-24
Inconsistent death count depending on simulation range


nielsenko:
I tried to run a simulation for Denmark with default data, first with the default simulation range, ie. until september 1st, and then one until January 1st 2021.

Checking the death count on the 31st of august in both runs, come up with widely different numbers, ie. 346 vs. 489.

Did the same for USA-New York and got: 7281 vs. 10245.

I suspect you suffer from significant numeric instability.

nnoll:
@nielsenko I suspect this is due to the reinterpolation of the mitigation plot onto the new time window and not due to numerical instability within the simulation. 

To test this, I ran Denmark twice with all default parameters, but with no mitigation, once for the default time window and once until January 1st 2021. Both times the number of deaths on August 31st was 66452. 

Thus we most likely need a more robust way of changing the mitigation curve when the time window updates.

nielsenko:
Good to hear you already have an hypothesis as to the cause. Good luck with your efforts.

albin-attentec:
I havs seen similar effects on the data-set for sweden, but only if i do one simulation after the other without reloading.
I belive this supports your theory @nnoll

nielsenko:
I have tried the preview (https://covid19-scenarios-git-fix-interpolation.covid19-scenarios.now.sh). It is much better but still not perfect.

Fx, running default parameters with CHE-Basel-Stadt, and then extending simulation time range until 1st of January 2021, gives a difference in accumulated deaths on the 31st of August of 1760 vs 1775.

nnoll:
@nielsenko I think with the current method that the mitigation is input by the user - i.e. discrete points that have to be re-interpolated onto a different time-window upon update - this is the best accuracy we will be able to achieve, especially when it directly modifies an exponential growth rate.

When Issue #14 is finally implemented the remainder of this problem will resolve as we won't have to guess at what to set values to upon time range updates. As this whole mitigation widget is being re-imagined near-term, we probably won't fix this small deviation in it's current form. Hope this helps!

PR #194, opened 2020-03-25
Change linear interpolation to spline

nnoll:
## Description

The function interpolateTimeSeries has used a linear interpolation that has numerical accuracy problems when attempting to interpolate the mitigation curve onto a different grid. This is exasperated by the fact that we display a spline interpolation.

## Related issues

Issue #187

The fundamental algorithm (which uses an interpolator to estimate the mitigation in between the sampled time points) and the mitigation curve.

## Testing

Needs to be done before I consider this done.


ivan-aksamentov:
@nnoll Could we use a library for the math? Just to make sure all the tricky corner cases are covered and to not maintain yet another library of matrix operations. We've removed `math.js` because it was not used, should we bring it back?

ivan-aksamentov:
Someone needs to help Nick with testing the new algorithm

abrie:
I'd like to help.

As expected, these changes break existing unit tests:

- TimeSeries.test.tsx:interpolateTimeSeries()
- model.test.ts:evolve()

Fixing those tests is simply a matter of updating the output snapshots.

In addition, there should be need a new test to verify that the new algorithm yields results expected by issue #187. I have a naive understanding of the problem, but maybe the test could be:

1. Run the model for one time series, and save the output.
2. Adjust the time series to be longer, run the model again, save the output.
3. Compare the two outputs, and verify that overlapping dates have the same values.

rneher:
My exploration of the result look sensible. reducing the length of the simulation gives the same dynamics as far as I can tell. This could be used as test: 
  
   * run a time series and save at mid-time
   * shrink the time series and compare results at the same time point.

Increasing the length of the time series is not ideal since the spline interpolation then extends into time ranges that weren't previously covered. Since splines are not strictly local, this can change the results also within the time range that was previously simulated. 

Another point that is worth considering is that we need to guard against negative values. 

image

I generated this by dragging two adjacent points to zero and then changing the number of points. 

An alternative fix for this would be to stick to the linear interpolation, interpolate the graph by lines instead of splines, and don't drop points when changing the time interval. This would generate messy graphs with different intervals and it is unclear what to do with the `number of points` box. But I am hesitant to sink too much effort into this as we plan to change this component soon.

nnoll:
@rneher regarding the extrapolation onto a larger date range - I kept the same basic logic as our previous linear interpolation, which is if we are asked to extrapolate beyond our current date range, then set it to the last value of the range. Is this not what you observe when you extend the date range?

I just pushed a fix to updateTimeSeries that guards against negative containment values.

nnoll:
OK I now use mathjs to perform the gauss jordan elimination.

ivan-aksamentov:
@nnoll I merged master into it and fixed tests (sort of)
This this ready?

nnoll:
@Ivan Yes, I haven't been able to break it yet and if we find an issue in deployment I'll fix it. I think merge when you can

PR #197, opened 2020-03-25
fix: add a command to close the disclaimer on each pages, update navigation test to not rely on language

Joluma:
## Description
With the introduction of i18n, the navigation bar now gets his content from the i18next. On the test, the content is now undefined because at the time we get the content of the links the translation is not made. This is not necessary since we only care about the link url being correct.
- update the test to rely on the url

Second issue is the introduction of the disclaimer popup, it breaks the tests because the area we ry to interact with are below the modal's overlay.
- add a command to close the popup automatically

## Related issues
 - Contributes to #19 

## Testing
run `yarn e2e` or `yarn e2e:dev`
covid-test

### DIFF
--- a/cypress/integration/navigation.spec.ts
+++ b/cypress/integration/navigation.spec.ts
@@ -6,7 +6,8 @@ context('The navigation bar', () => {
   const navLinks = Object.entries(links)
 
   beforeEach(() => {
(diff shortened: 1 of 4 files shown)


ivan-aksamentov:
@Joluma Thanks for updating this,  Joël!

PR #210, opened 2020-03-25
internationalize number display

### LINKED ISSUE #198: Internationalize and humanize numbers (at the same time)
# 🙋 Feature Request

Now that the numbers are displayed in a human-readable form, would be nice to also make them internationalized.

## 😯 Describe the feature

It is hard to describe all the requirements specifically, without knowing the languages. This should be covered by a library.

## 💻 Examples

Here is an non-exhaustive list of requirements:
 - [ ] in some locales comma is used instead of the dot as a decimal separator
 - [ ] in some locales thousands (or even tenth of thousands) are space-separated
 - [ ] exponent suffixes "k", "M"  should ideally be translated, (e.g. "k" is "Tnd." in German and "тыс." in Russian)

## 💁 Possible Solution

A mature library should be used. Either D3.js  currently used or number formatting, or react-i18next used for i18n or both.

## Related:

 - This is a continuation of https://github.com/neherlab/covid19_scenarios/issues/158  implemented in https://github.com/neherlab/covid19_scenarios/pull/196 (except for i18n part)
 - https://github.com/neherlab/covid19_scenarios/issues/56

gj262:
## Related issues and PRs

Fixes #198 

## Description

- Always internationalize numbers
- Optionally 'humanize' them
- Keep the total number rounding in the outcomes table when numbers are non humanized

Number display

## Testing

Check numbers on axis and tooltips.

@rcbevans can you check this? Thanks!

### DIFF
--- a/src/components/Main/Results/OutcomeRatesTable.tsx
+++ b/src/components/Main/Results/OutcomeRatesTable.tsx
@@ -10,6 +10,8 @@ import { AlgorithmResult } from '../../../algorithms/types/Result.types'
 
 import { SeverityTableRow } from '../Scenario/SeverityTable'
 
+import { numberFormatter } from '../../../helpers/numberFormat'
+
 export interface TableProps {
   showHumanized?: boolean
   result?: AlgorithmResult
@@ -17,15 +19,15 @@ export interface TableProps {
 }
 
 const percentageFormatter = (v: number) => d3.format('.2f')(v * 100)
-const humanizeFormatter = d3.format('.5s')
-const decimalFormatter = d3.format('d')
 
 export function OutcomeRatesTable({ showHumanized, result, rates }: TableProps) {
-  const { t } = useTranslation()
+  const { t, i18n } = useTranslation()
 
   if (!result || !rates) {
     return null
   }
+
+  const formatNumber = numberFormatter(i18n.language, !!showHumanized, true)
 
   /*
   // FIXME: This looks like a prefix sum. Should we use `Array.reduce()` or a library instead?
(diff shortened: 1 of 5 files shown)


src/helpers/numberFormat.ts
```diff
@@ -0,0 +1,6 @@
+export function numberFormatter(language: string | undefined, humanize: boolean, round: boolean) {
+  return new Intl.NumberFormat(language, {
+    ...(humanize ? { notation: 'compact', compactDisplay: 'short' } : {}),
```
rcbevans: Thanks for tidying this up, was a bit lazy repeating the formatters in the various files.

gj262: Yeah, I was going the lazy route too but Intl.Number... lines are so much longer than d3.format and I had to update the &*(^ typescript definitions so that pushed me to consolidation :)

src/components/Main/Results/OutcomeRatesTable.tsx
```diff
@@ -62,7 +64,7 @@ export function OutcomeRatesTable({ showHumanized, result, rates }: TableProps)
     Math.max(...result.deterministicTrajectory.map((x) => x.critical.total + x.overflow.total)),
   )
 
-  const totalFormatter = (value: number) => (showHumanized ? humanizeFormatter(value) : decimalFormatter(value))
+  const totalFormatter = (value: number) => formatNumber(value)
```
rcbevans: This is causing some substantial inaccuracy of the tabular results.  e.g for US simulation, the total dead is 1.67M, that's being rounded all the way up to 2M.  Would it be better to round to a couple of decimal places?  330k difference is quite big.

rcbevans: One option could be to change the rounding bool, to a number indicating number of decimal places, which would keep the same functionality possible, with more flexibility

gj262: Ikes. Thats not good, I'll re-open the issue. Thanks for spotting it.

PR #213, opened 2020-03-25
do not round outcome rates when compacting them

### LINKED ISSUE #198: Internationalize and humanize numbers (at the same time)

gj262:
## Related issues and PRs

Fix #198 & resolves #210 regression where 1667132.485331003 became 2M.

## Description

Do not round the values in the outcome rates table if they are being displayed in a humanly readable compact form. When the human form causes truncation use 2 fraction digits. e.g. 1667132.485331003 becomes 1.67M for locale "en".

Do round them otherwise.

And always i18n them.

Number display

## Testing

Look at the numbers!

### DIFF
--- a/src/helpers/numberFormat.ts
+++ b/src/helpers/numberFormat.ts
@@ -1,6 +1,6 @@
-export function numberFormatter(language: string | undefined, humanize: boolean, round: boolean) {
+export function numberFormatter(language: string | undefined, compact: boolean, round: boolean) {
   return new Intl.NumberFormat(language, {
-    ...(humanize ? { notation: 'compact', compactDisplay: 'short' } : {}),
+    ...(compact ? { notation: 'compact', compactDisplay: 'short', maximumFractionDigits: 2 } : {}),
     ...(round ? { maximumFractionDigits: 0 } : {}),
   }).format
 }
--- a/src/helpers/numberFormat.ts
+++ b/src/helpers/numberFormat.ts
@@ -1,6 +1,14 @@
-export function numberFormatter(language: string | undefined, compact: boolean, round: boolean) {
-  return new Intl.NumberFormat(language, {
-    ...(compact ? { notation: 'compact', compactDisplay: 'short', maximumFractionDigits: 2 } : {}),
-    ...(round ? { maximumFractionDigits: 0 } : {}),
-  }).format
+import * as d3 from 'd3'
+
+export function numberFormatter(humanize: boolean, round: boolean) {
+  let formatString = ',.2f'
+  if (humanize && round) {
+    throw new Error('not implemented')
+  } else if (humanize) {
+    formatString = ',.3~s'
+  } else if (round) {
+    formatString = ',.0f'
+  }
+
+  return d3.format(formatString)
 }
--- a/src/components/LanguageSwitcher/index.tsx
+++ b/src/components/LanguageSwitcher/index.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react'
 import { Dropdown, DropdownToggle, DropdownMenu, DropdownItem } from 'reactstrap'
 import i18next from 'i18next'
+import * as d3 from 'd3'
 
 import langs, { Lang as LangType } from '../../langs'
 
@@ -46,6 +47,7 @@ export default function LanguageSwitcher() {
               i18next.changeLanguage(langs[key].lang, () => {
                 localStorage.setItem('lang', langs[key].lang)
               })
+              d3.formatDefaultLocale(langs[key].d3Locale)
             }}
           >
             <Lang lang={langs[key]} />
--- a/src/components/LanguageSwitcher/index.tsx
+++ b/src/components/LanguageSwitcher/index.tsx
@@ -2,7 +2,7 @@ import React, { useState } from 'react'
 import { Dropdown, DropdownToggle, DropdownMenu, DropdownItem } from 'reactstrap'
 import i18next from 'i18next'
 
-import langs from '../../langs'
+import langs, { Lang as LangType } from '../../langs'
 
 import './LanguageSwitcher.scss'
 
@@ -24,7 +24,7 @@ export function getCurrentLang(): string {
  * Flag next?
  * @param lang
  */
-function Lang({ lang }: { lang: Lang }) {
+function Lang({ lang }: { lang: LangType }) {
   return <span className="language-switcher-lang">{lang.name}</span>
 }
 
--- a/src/components/Main/Results/AgeBarChart.tsx
+++ b/src/components/Main/Results/AgeBarChart.tsx
@@ -23,13 +23,13 @@ export interface SimProps {
 }
 
 export function AgeBarChart({ showHumanized, data, rates }: SimProps) {
-  const { t: unsafeT, i18n } = useTranslation()
+  const { t: unsafeT } = useTranslation()
 
   if (!data || !rates) {
     return null
   }
 
-  const formatNumber = numberFormatter(i18n.language, !!showHumanized, false)
+  const formatNumber = numberFormatter(!!showHumanized, !showHumanized)
 
   const t = (...args: Parameters<typeof unsafeT>) => {
     const translation = unsafeT(...args)
--- a/src/components/Main/Results/DeterministicLinePlot.tsx
+++ b/src/components/Main/Results/DeterministicLinePlot.tsx
@@ -77,9 +77,9 @@ function legendFormatter(enabledPlots: string[], value: string, entry: any) {
 }
 
 export function DeterministicLinePlot({ data, userResult, logScale, showHumanized, caseCounts }: LinePlotProps) {
-  const { t, i18n } = useTranslation()
+  const { t } = useTranslation()
 
-  const formatNumber = numberFormatter(i18n.language, !!showHumanized, false)
+  const formatNumber = numberFormatter(!!showHumanized, !showHumanized)
 
   const [enabledPlots, setEnabledPlots] = useState(Object.values(DATA_POINTS))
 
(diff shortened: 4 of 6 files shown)


gj262:
Hi @rcbevans can you give this a look when you can? Thanks.

src/helpers/numberFormat.ts
```diff
@@ -1,6 +1,6 @@
-export function numberFormatter(language: string | undefined, humanize: boolean, round: boolean) {
+export function numberFormatter(language: string | undefined, compact: boolean, round: boolean) {
```
rcbevans: Unless people are familiar with the Intl.NumberFormat API they may not know that `compact` corresponds to humanizing the value.

This looks fine though.  I like the flexibility.

rcbevans:
> This actually doesn't look like it's humanizing anymore.

image

gj262:
> This actually doesn't look like it's humanizing anymore.

Hi Rich, I cannot reproduce that problem. I tried changing parameters, using autosave. Do you recall what steps you went through?
Thanks,
-Gavin

rcbevans:
I don't recall doing any steps in particular.  I just opened up the test site (in Firefox), set population scenario to USA and then ran the simulation.

None of the values in results were humanized; toggling the humanize results toggle made no difference.

rcbevans:
For reference, I'm using Firefox 74.0 (64 bit) on MacOS Catalina.  Will try other browsers.

rcbevans:
I see the same in Safari.

gj262:
thanks @rcbevans I see it in FF also. taking a look.

gj262:
So Firefox and Safari do not support Intl.NumberFormat as well as MDN had me believe. So I'm taking a look at the numeral JS lib. As per i18next:

https://www.i18next.com/translation-function/formatting#formatting
http://numeraljs.com/

This should provide internationalization and 'humanization' and I can add some unit tests to make sure :)

ivan-aksamentov:
@rcbevans  @gj262 Thanks for the investigation, folks

gj262:
> http://numeraljs.com/

*sigh* numeralJS blows up if I change locale en->de :(

next up: react-intl

ivan-aksamentov:
@gj262 At some point we used `d3-format`. Looks like it has i18n features. 
https://github.com/d3/d3-format
If nothing else works, do you think this would be suitable?

rcbevans:
Yeah originally I implemented it using d3-format, I just didn't add the internationalization in.  It does support it though.

https://d3-wiki.readthedocs.io/zh_CN/master/Localization/

You just need to call `d3.locale(<Lang>).numberFormat(<formatting>)`. As opposed to `d3.numberFormat(<formatting>)` as I was previously.

gj262:
yeah. I'm looking at that. (react-intl uses Intl.NumberFormat and the polyfills I can find do not support humanizing).

I'm trying to come up with a good way to load d3 locales and match with the language switcher. What I have is using lang.ts:

```
import enLocale from 'd3-format/locale/en-US.json'
import deLocale from 'd3-format/locale/de-DE.json'
...
const langs: { [key: string]: Lang } = {
  en: {
    lang: 'en',
    name: 'english',
    d3Locale: enLocale as FormatLocaleDefinition,
  },
...
```

which is not that elegant IMO.

ivan-aksamentov:
@gj262 are locales large? If it can be async, maybe we could dynamic `import()` and code-split it? Except for English.

gj262:
ok @rcbevans I think I've got it this time. Thanks for your previous diligence in catching my big number and browser bugs. I was more diligent myself with testing this time! 

I ran a matrix of tests:

- Chrome, FireFox, Safari
- All Languages
- Humanized and Not
- The default Population and USA (for large number testing).

For each one I eyeballed axis labels, hover values and table values and all seem OK. 

I plan to add a unit test...

Loading the d3 locales seems klunky to me but it works.

Minor issue: In non humanized mode with large numbers the Y axis hides the top 100M label and the distribution graph legends are over the labels. I'm not sure this is worth addressing urgently as humanized mode is the default.

gj262:
> @gj262 are locales large? If it can be async, maybe we could dynamic `import()` and code-split it? Except for English.

They are v. small. e.g. 
```
{
  "decimal": ".",
  "thousands": ",",
  "grouping": [3],
  "currency": ["$", ""]
}
```

ivan-aksamentov:
@gj262 I see. So we won't have the replacements for "k" and "M" :(

@rcbevans Could you please take a look also?

gj262:
> @gj262 I see. So we won't have the replacements for "k" and "M" :(

Yeah, d3 only internationalizes separators as far as I can tell.

Intl.NumberFormat might work for chrome and then d3 for browsers that don't properly support it. But it is not perfect IMO. e.g.

```
new Intl.NumberFormat('de-DE', { notation: "compact" , compactDisplay: "short", maximumFractionDigits: 2 }).format(123456)
"123.456"
new Intl.NumberFormat('de-DE', { notation: "compact" , compactDisplay: "short", maximumFractionDigits: 2 }).format(123456789)
"123,46 Mio."
```

NumeralJS looked promising but I could not figure out in the short term why it was crashing on locale switch.

rcbevans:
> > @gj262 I see. So we won't have the replacements for "k" and "M" :(
> 
> Yeah, d3 only internationalizes separators as far as I can tell.
> 
> Intl.NumberFormat might work for chrome and then d3 for browsers that don't properly support it. But it is not perfect IMO. e.g.
> 
> ```
> new Intl.NumberFormat('de-DE', { notation: "compact" , compactDisplay: "short", maximumFractionDigits: 2 }).format(123456)
> "123.456"
> new Intl.NumberFormat('de-DE', { notation: "compact" , compactDisplay: "short", maximumFractionDigits: 2 }).format(123456789)
> "123,46 Mio."
> ```
> 
> NumeralJS looked promising but I could not figure out in the short term why it was crashing on locale switch.

Do you have a branch with Numeral?  Definitely looks like that is the most promising library for i18n humanization of values.

I'm happy to take a look to try and help figure out the crash.

rcbevans:
So I have a branch where I'm using numeral for internationalized formatting, and it's working across browsers.  It doesn't seem like the default abbreviations for the humanized numbers are particularly good though.

For example the [fr locale](https://github.com/adamwdraper/Numeral-js/blob/master/locales/fr.js) has
```
        abbreviations: {
            thousand: 'k',
            million: 'm',
            billion: 'b',
            trillion: 't'
        },
```

However, [fr-ca](https://github.com/adamwdraper/Numeral-js/blob/master/locales/fr-ca.js) has 
```
        abbreviations: {
            thousand: 'k',
            million: 'M',
            billion: 'G',
            trillion: 'T'
        },
```

Numeral doesn't even offer an "en" locale, only "en-gb", "en-aus" and "en-za".

rcbevans:
I created a PR.  You may have been seeing crashes on language switch if you attempted to either, load a locale which is not loaded into numeral, i.e "en", rather than "en-gb", "en-aus", or "en-za", or if the locale itself, whilst available, hadn't been loading using an appropriate `import 'numeral/locales/<locale>'` statement.

Either way, feel free to use my PR, or elements of it.

If we are not happy with the abbreviations provided by the numeral library locales we can provide our own, based on user needs, or even better, contribute them back to the numeraljs codebase.

rcbevans:
> > @gj262 I see. So we won't have the replacements for "k" and "M" :(
> 
> Yeah, d3 only internationalizes separators as far as I can tell.
> 
> Intl.NumberFormat might work for chrome and then d3 for browsers that don't properly support it. But it is not perfect IMO. e.g.
> 
> ```
> new Intl.NumberFormat('de-DE', { notation: "compact" , compactDisplay: "short", maximumFractionDigits: 2 }).format(123456)
> "123.456"
> new Intl.NumberFormat('de-DE', { notation: "compact" , compactDisplay: "short", maximumFractionDigits: 2 }).format(123456789)
> "123,46 Mio."
> ```
> 
> NumeralJS looked promising but I could not figure out in the short term why it was crashing on locale switch.

It could be useful to take your Intl.NumberFormat implementation on chrome and get a collection of abbreviations, as per Chrome implementation, then create numeral.js locales based on those.

gj262:
Could use command line node:
```
[gavin@pelican covid19_scenarios (eslintingII)]$ node
Welcome to Node.js v13.11.0.
Type ".help" for more information.
> new Intl.NumberFormat('de-DE', { notation: "compact" , compactDisplay:
"short", maximumFractionDigits: 2 }).format(123456)
'123.456'
> new Intl.NumberFormat('de-DE', { notation: "compact" , compactDisplay:
"short", maximumFractionDigits: 2 }).format(123456789)
'123,46 Mio.'
```

BTW I think I'll close this in favor of #241. What do you think?

ivan-aksamentov:
@rcbevans @gj262 Can `Intl.NumberFormat` be pollyfilled?
It is a shady story. Every single place on the internet says that `Intl.NumberFormat` is supported since Firefox 29. Why it does not work in our case?

gj262:
Hi @ivan-aksamentov the polyfills that I could find do not implement the notation:compact feature that I had hoped would fulfill the feature request. And then as @rcbevans found browsers have incomplete support Intl.NumberFormat. I think #240 will meet the immediate need and can be improved on if the abbreviations are not quite right. But I agree that it is shady that this is not simply implemented *somewhere*.

gj262:
This will be fixed with #241.

Issue #31, opened 2020-03-18
Link to previous scenarios


nnoll:
It would be great for people to link to custom configured scenarios via a dynamically generated URL. This would allow people to easily share their results.

ivan-aksamentov:
This will either require:

  * encoding the entire scenario state in URL params

  * a backend to store the results (which is not impossible with the community efforts)

Similar alternative solution is "import". In this case users can share the param file and/or results file(s).

Related:
 - https://github.com/neherlab/covid19_scenarios/issues/11

brunorzn:
I had a look at the "encoding URL" option, which is totally doable.

It will require some work on the "initial state" part, but nothing too complicated though.
The URL will be a bit long, but is it a problem ?
(see  http://graphql.org/swapi-graphql?query=%7B%0A%20%20allVehicles%20%7B%0A%20%20%20%20edges%20%7B%0A%20%20%20%20%20%20node%20%7B%0A%20%20%20%20%20%20%20%20vehicleClass%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A for an example)

The fastest and not most optimised way to do so would probably be to serialise all parameters as JSON and encode them in the URL. 

The main drawback I see with this approach is that it won't necessary be retro-compatible (a URL saved 2 weeks ago probably won't work if breaking changes have been made).
But again, is it really an issue ?

I suggest to take this approach, and see if we do need a backend after all (in that case, we will still have the required "init state" part from above done).

What do you think ?

nnoll:
Regarding the reto-compatibilty I don't think this is a concern of huge importance -- the ability for people to share results now would take precedence over the ability to reference results from two weeks ago. The import/export parameters of Issue #11 could solve this use case.

brunorzn:
#53

ivan-aksamentov:
@brunorzn Looks like it works! :)

I merged #53, but I noticed that it does not restore the names of the scenarios (in dropdowns in card headers). That needs to be fixed, I think.

Additionally, we are using `connected-react-router`, and I think bypassing it may break it. Or at least all the places that rely or will be relying on it will not receive the updated state.
Also, router should parse params for you, if you'd be willing to trasition from urlencoded JSON to proper URL params (e.g.`?param=value&param2=value`)

Let me know if you'd like to work on a followup.

brunorzn:
@ivan-aksamentov 

cool for the merge :)

[...]

PR #214, opened 2020-03-25
Abstract LocalStorage and persist user toggles

rcbevans:
## Related issues and PRs

N/A

## Description

The goal was to abstract setting/getting LocalStorage, for mockability in tests, to centralize local storage item keys, and add persistence of UI toggles, such as autorun, show humanized results, and logarithmic graphs.

User toggle settings across page refreshes.

## Testing
Open the site, toggle a UI setting, such as auto run, log graphs and show humanize.  Refresh the page, observe that your UI preference is restored.


src/helpers/localStorage.ts
```diff
@@ -0,0 +1,30 @@
...
+  SUPPRESS_DISCLAIMER: 'suppress-disclaimer',
+}
+
+function set<T>(key: string, value: T): void {
+  localStorage.setItem(key, typeof value !== 'string' ? JSON.stringify(value) : value)
+}
+
+function get<T>(key: string): T | null {
+  const value = localStorage.getItem(key)
+  try {
```
rcbevans: I'm not aware of any way in Typescript to assert the typeof T at runtime to prevent attempting to JSON.parse something unparseable, like a normal string, for example the language.

If someone has a suggestion as to how the type safety here can be improved, would love to learn how!

ivan-aksamentov: Yep, TS does not have any runtime, so runtime validation should be performed.
We validate form params declaratively with `yup`:
https://github.com/neherlab/covid19_scenarios/blob/master/src/components/Main/validation/schema.ts

So this part might be shared. And I thing URL serialization/deserialization folks in https://github.com/neherlab/covid19_scenarios/issues/31 may add `io-ts`. Which is very typescript-y and awesome.

But something imperative like a good old `validate.js` or in simple cases even `lodash` should do.

PR #118, opened 2020-03-22
Fix scripts for Windows environment

mickeypearce:
## Description
Fixing package.json scripts for Win environment

## Related issues
- Fixes #38

PR #227, opened 2020-03-26
Add run-s to sequentially run dev scripts

mickeypearce:
## Related issues and PRs
#38 and #118 

## Description
Shell scripts with ";" between commands don't work properly on Windows...
Adding a dependency: "npm-run-all" to solve the problem of running scripts sequentally without exiting on error.

All complex scripts in package.json


ivan-aksamentov:
@mickeypearce Thanks!

PR #235, opened 2020-03-26
Restructure the internal simulation data to allow for better nomenclature.

nnoll:
## Related issues and PRs

None

## Description

The simulation states have grown to accommodate a more realistic description of the pandemic. Our terminology within the simulation and its corresponding data structure has not been updated accordingly. 

This refactoring addresses this and should make the terms in the equation, as well as the connection of states, easier to follow logically.

The internal algorithm

## Testing

I expect tests to break as I changed the internal data structure.


src/algorithms/model.ts
```diff
@@ -140,207 +144,261 @@ export function initializePopulation(
...
+    pop.cumulative.fatality[k] = 0
     if (i === Math.round(Object.keys(ages).length / 2)) {
-      pop.susceptible[k] -= numCases
-      pop.infectious[k] = (0.3 * numCases)
-      const e           = (0.7 * numCases) / pop.exposed[k].length;
-      pop.exposed[k]    = [e, e, e];
+      pop.current.susceptible[k] -= numCases
+      pop.current.infectious[k] = 0.3 * numCases
+      const e = (0.7 * numCases) / pop.current.exposed[k].length
+      pop.current.exposed[k] = [e, e, e]
```
rneher: if we can make this not imply the length of the array, that would be great!

nnoll: Updated the branch. The value is still hardcoded in one spot  but I think this is OK as everything else will follow from that.

PR #157, opened 2020-03-23
Feat: Responsive tooltips

miksa1987:
Here is my proposition on responsive tooltips, as discussed here: [#115](https://github.com/neherlab/covid19_scenarios/issues/115)

Originally plan was to make a custom component for tooltips, but I had to abandon that as of now as I couldn't get it to work. However, I came up with a kind of responsive y position calculator, and also changed the z-index of tooltips to 999 so other charts do not cover them when hovering. 

Please let me know what you think could be improved, maybe the position calculation could be tweaked a little.

PR #237, opened 2020-03-26
Feat: Responsive tooltips

miksa1987:
All right, here's a fixed PR discussed here: [#157](https://github.com/neherlab/covid19_scenarios/pull/157)

There was some weirdness with rebasing. So this PR should only deal with my changes, which include:

- Custom ResponsiveTooltipContent component, which will divide in two columns on mobile for better readability.
- Repositioning of the chart Tooltip component when in single column mode.
- When clicking on chart in single column mode, the page scrolls the chart to top, so that repositioned Tooltip is wholly visible.

@ivan-aksamentov if you're happy with this you can merge.

### DIFF
--- a/src/components/Main/Results/DeterministicLinePlot.tsx
+++ b/src/components/Main/Results/DeterministicLinePlot.tsx
@@ -8,6 +8,9 @@ import { AlgorithmResult, UserResult } from '../../../algorithms/types/Result.ty
 import { EmpiricalData } from '../../../algorithms/types/Param.types'
 import { numberFormatter } from '../../../helpers/numberFormat'
 
+import { calculatePosition, scrollToRef } from './chartHelper'
+import { ResponsiveTooltipContent } from './ResponsiveTooltipContent'
+
 import './DeterministicLinePlot.scss'
 
 const ASPECT_RATIO = 16 / 9
@@ -81,6 +84,8 @@ export function DeterministicLinePlot({ data, userResult, logScale, showHumanize
 
   const formatNumber = numberFormatter(!!showHumanized, false)
   const formatNumberRounded = numberFormatter(!!showHumanized, true)
+
+  const chartRef = React.useRef(null)
 
   const [enabledPlots, setEnabledPlots] = useState(Object.values(DATA_POINTS))
 
@@ -209,11 +214,15 @@ export function DeterministicLinePlot({ data, userResult, logScale, showHumanize
           }
 
           const height = Math.max(500, width / ASPECT_RATIO)
+          const tooltipPosition = calculatePosition(height)
 
           return (
             <>
               <h5>{t('Cases through time')}</h5>
+              
+              <div ref={chartRef} />
               <ComposedChart
+                onClick={() => scrollToRef(chartRef)}
                 width={width}
                 height={height}
                 data={plotData}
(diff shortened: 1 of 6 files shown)


rcbevans:
Tooltip values do not appear to be internationalized/humanized:

image

src/components/Main/Results/ResponsiveTooltipContent.tsx
```diff
@@ -0,0 +1,55 @@
...
+  if (active) {
+    return (
+      <div className="responsive-tooltip-content-base">
+        <strong>{formattedLabel}</strong>
+        <div className="responsive-tooltip-content">
+          <div>
+            {left.map((item, idx) => <ToolTipContentItem key={idx} name={item.name} value={item.value} color={item.color} />)}
+          </div>
+          <div>
+            {right.map((item, idx) => <ToolTipContentItem key={idx} name={item.name} value={item.value} color={item.color} />)}
```
rcbevans: I believe this is the `item.value` that needs to be appropriately formatted for display.

src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -230,7 +239,11 @@ export function DeterministicLinePlot({ data, userResult, logScale, showHumanize
                   tickCount={7}
                 />
                 <YAxis scale={logScaleString} type="number" domain={[1, 'dataMax']} tickFormatter={yTickFormatter} />
-                <Tooltip formatter={tooltipFormatter} labelFormatter={labelFormatter} />
+                <Tooltip 
+                  formatter={tooltipFormatter}
```
rcbevans: You need to pass this tooltipFormatter to your ResponsiveTooltipContent and format the right hand value as necessary.  I don't think the tooltip is using the formatter, since custom content takes priority.

miksa1987:
@ivan-aksamentov number formatting should work now as intended. The ResponsiveTooltipContent now formats the values with the formatter if passed to it. 

However, I'm wondering about all the parameters tooltipFormatter needs, if it only returns the formatted value. Should it be simplified to take only the value and format that, or is there a thought on expanding it's capabilities that I should take into account on my component?

src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -230,7 +239,11 @@ export function DeterministicLinePlot({ data, userResult, logScale, showHumanize
                   tickCount={7}
                 />
                 <YAxis scale={logScaleString} type="number" domain={[1, 'dataMax']} tickFormatter={yTickFormatter} />
-                <Tooltip formatter={tooltipFormatter} labelFormatter={labelFormatter} />
+                <Tooltip 
+                  formatter={tooltipFormatter} 
+                  position={tooltipPosition}
```
gj262: Can you pass `labelFormatter={labelFormatter}` also so that `<ResponsiveTooltipContent />` does not have to guess as to label type i.e. date -v- age.

src/components/Main/Results/ResponsiveTooltipContent.tsx
```diff
@@ -0,0 +1,58 @@
...
+      {name}
+      <div className="responsive-tooltip-content-placeholder" />
+      {value}
+    </div>
+  )
+}
+
+export function ResponsiveTooltipContent({ active, payload, label, formatter }: TooltipContentProps) {
+  const formattedLabel = Number(label) > 10000 ? new Date(label).toISOString().slice(0, 10) : label
+  const formatNumber = formatter
```
gj262: could use labelFormatter here (I think)

miksa1987:
> Looks great. See comment re: label formatting.

Thanks for noting that. I updated my component to use the labelFormatter and passed it in DeterministicLinePlot. I also rebased the whole thing.

ivan-aksamentov:
Excellent job!

Issue #234, opened 2020-03-26
Improve positioning and movement of the results graph tooltip


ivan-aksamentov:
# 🙋 Feature Request

Currently the tooltip positioning is very chaotic. The smooth animations sometime work fine, sometimes the tooltip jerks and twitches like crazy.

![Peek 2020-03-26 18-26](https://user-images.githubusercontent.com/9403403/77680312-12d4cb80-6f94-11ea-93e0-b7103a44f269.gif)

## 😯 Describe the feature

This feature require some tweaking and checking the overall UX. Cannot give exact TODOs, but this would be nice to include:
 - [ ] prevent any jerkiness and twitching from happening. Partially addressed in https://github.com/neherlab/covid19_scenarios/pull/239
 - [ ] make tooltip follow the mouse without too much of a lag (or a little lag just to highlight the movement of the vertical indicator line on the plot)
 - [ ] disable as much animations as possible, to improve compatibility on different browsers and devices
 - [ ] z-index: prevent tooltip to be obscured by the text on  the graph below when overlaps
 - [ ] these numbers is the most important part of the application. Why are they in the tooltip?
 - [ ] be creative, try different things and approaches that lead to better UX

**NOTE**: we are talking specifically about tooltip and nothing else

## Related

 - https://github.com/neherlab/covid19_scenarios/issues/233
 - https://github.com/neherlab/covid19_scenarios/pull/157
 - https://github.com/neherlab/covid19_scenarios/pull/239

kmid5280:
@ivan-aksamentov A few questions:
- Which animations are you currently aware are happening within the application?
- Can you provide steps to reproduce the issue with the z-index, where the text on the graph obscures the tooltip?
- You mentioned numbers in the tooltip. Which numbers are you referring to and what changes would you prefer?

ivan-aksamentov:
@kmid5280 

> Which animations are you currently aware are happening within the application?

This issues is specifically about results plot tooltip. And I don't know which animations are enabled there. I think the default ones that come from the plotting library.

> Can you provide steps to reproduce the issue with the z-index, where the text on the graph obscures the tooltip?

You just move mouse around until the tooltip is positioned such that it goes outside of the plot on its bottom part:

![Screenshot from 2020-03-27 00-44-28](https://user-images.githubusercontent.com/9403403/77707002-344dab80-6fc4-11ea-936a-ab4b6f07bb84.png)

Because tooltip is quite chaotic, I don't really have exact steps.

[...]

PR #239, opened 2020-03-26
Jittery chart fix

kmid5280:
## Related issues and PRs
Contributes to #234 

## Description
Make tooltip positioning in results chart smoother when mouse is moved over graph.

New throttleDelay value added in ComposedChart component in src/components/Main/Results/DeterministicLinePlot.tsx

## Testing
Move mouse over tooltips in graph and see if desired smoothness is achieved.

### DIFF
--- a/src/components/Main/Results/DeterministicLinePlot.tsx
+++ b/src/components/Main/Results/DeterministicLinePlot.tsx
@@ -214,6 +214,7 @@ export function DeterministicLinePlot({ data, userResult, logScale, showHumanize
                 width={width}
                 height={height}
                 data={plotData}
+                throttleDelay={75}
                 margin={{
                   left: 15,
                   right: 15,


kmid5280:
Some issues that reference general chart jitteriness in this plugin: https://github.com/recharts/recharts/issues/1499 https://github.com/recharts/recharts/issues/1794

ivan-aksamentov:
@kmid5280 Thanks!

I see the difference, I think. It does not jump as much as before. But the difference from master is not huge.

Tooltip still jumps somewhat when you hover and mess around the confirmed cases "dots" (in the bottom-left corner of the plot)

I think we need a deeper dig into the library's docs and its components.
I'd like to also address other points in the issue, so I keep it open.

kmid5280:
@ivan-aksamentov I addressed this issue by setting a value for throttleDelay in the ComposedChart component, line 217. Decreasing that value will decrease the delay, but will increase the jitter if it's too low. So there is likely a happy medium somewhere.

leeoniya:
@ivan-aksamentov @kmid

depending on the types of charts you need, maybe evaluate [uPlot](https://github.com/leeoniya/uPlot) (shameless plug :D)

PR #240, opened 2020-03-26
Format numbers using numeraljs

rcbevans:
## Related issues and PRs
Alternative to https://github.com/neherlab/covid19_scenarios/pull/213
Fixes https://github.com/neherlab/covid19_scenarios/issues/198 

## Description

Use numeraljs to allow internationalizing and human formatting values simultaneously

Screenshot of spanish language with humanized formatting

Result charts and tables.

## Testing
Switch language with a scernario ran, and humaized results enabled and validate that axis, tooltips and tables are showing appropriately humanized, and internationalized values.


src/i18n.ts
```diff
@@ -32,4 +32,12 @@ i18n
     },
   })
 
+i18n.on('languageChanged', lang => {
+  numeral.locale(langs[lang].numeralLocale)
+})
+
```
gj262: 🆒

rcbevans:
Gah looks like my local branch got b0rked somehow...

$ git submodule update --init --recursive
Submodule 'data' (https://github.com/neherlab/covid19_scenarios_data) registered for path 'covid19_scenarios_data'
fatal: No url found for submodule path 'data' in .gitmodules

gj262:
It looks like the formatting is always adding 3 fractional digits:
Screen Shot 2020-03-26 at 3 56 09 PM

src/i18n.ts
rcbevans: > It looks like the formatting is always adding 3 fractional digits:
> Screen Shot 2020-03-26 at 3 56 09 PM

Yes it is, we can easily change it to be more or less, depending what we want, it's controlled by the format string in numberFormatter.

'0.000a' is the default non-rounded, humanized string, with .000 being 3 decimal places.

rcbevans: i can have a lot and see if there's a way of trimming extra 0s.

rcbevans: Ah yes, there is.  `[0000]`

src/locales/de/index.ts
```diff
@@ -1,3 +1,4 @@
 import * as translation from './translation.json'
+import 'numeral/locales/de'
```
gj262: per @ivan-aksamentov request in the #198 "exponent suffixes "k", "M" should ideally be translated, (e.g. "k" is "Tnd." in German and "тыс." in Russian)".  The import would need to change to modify the abbreviations in:
https://github.com/adamwdraper/Numeral-js/blob/master/locales/de.js

src/i18n.ts
rcbevans: image

src/locales/de/index.ts
rcbevans: Yep.  I don't have a list of all the abbreviations wanted, else I'd do it myself.

rcbevans:
@gj262 github freaked out when I recreated and forced push the branch, so I recreated the PR, #241

Issue #11, opened 2020-03-15
❗[IMPORTANT] Import scenarios


pinanunes:
I believe you might have that on the roadmap since there is an option to export.
It would be great to name and compare scenarios with simple KPI.

rneher:
Yes, import is on the map. But what is KPI?

pinanunes:
Sorry. Key Process Indicators or main outcomes of the scenarios that can be the Total/peak figures

ivan-aksamentov:
Hello @pinanunes ,

This is something currently in the works on the branch [`feat/compare-rebased-2`](https://github.com/neherlab/covid19_scenarios/tree/feat/compare-rebased-2). As you see, the branch had a number of rebases already,  we aren't quite sure yet how exactly to display the results being uploaded/compared.

For your use-case: 

 * What kind of uploading/comparison workflow you imagine would be the most convenient?

 * How do you think it is better to display the uploaded results?

We appreciate any advice.

pinanunes:
Hello @ivan-aksamentov ,

I would say we have 3 major outcomes we want to compare between scenarios either in absolute or relative numbers to a basic scenario:
1-Number of deaths 
2- Number of cases
3- Peak value of critical cases

One option would be to save server-side on the session several scenarios with a "Add to compare" button where you would name the scenario. On a different tab, you would have those values as table format, scenarios as rows, indicators as columns and a graph with fewer series than on main page (cases, deaths), color for the scenario, shape (straight, dashed) for the indicator.
Eventually, instead of using a table of numbers, you can use a table with valueboxes (for example in shiny https://rstudio.github.io/shinydashboard/structure.html#valuebox) with colors coded as difference to first scenario or the scenario number. You can have a dropdown box to select the indicator you want to compare on top. 

Another (or complementary) approach would be to have this compare tab and add scenarios by importing the json files (and running each one).


Besides scenario comparison, adding an import button on the main page would be useful. Can also be on the 1st dropdown box an option to import that on select would open an overlay window to drag and drop or browse local folders for the the .json file.

Check Imperial College's scenario exploration here: https://www.imperial.ac.uk/media/imperial-college/medicine/sph/ide/gida-fellowships/Imperial-College-COVID19-NPI-modelling-16-03-2020.pdf

It depends much on the balance you want to have between presenting much information on one page or making it visually appealing

Count me in for comments suggestions

loehndorf:
[...]

PR #243, opened 2020-03-27
#101/Add type for case_counts.json, and a transformer into EmpiricalData type.

abrie:
## Related issues and PRs

#101 

## Description

The goal is to remove type warnings from the application. Specifically this one:

image

`countryCaseCountData` is imported from JSON. This PR implements a transformer to convert the imported, untyped, `case_counts.json` to typed EmpiricalData.

At this point this PR is just a proposal. Maybe there is a better way to do this?

All parts of the application that use import data from  `assets/data/case_counts.json`

## Testing

PR includes unit tests to verify behavior: `src/algorithms/test/JsonToTypes.test.ts`


src/algorithms/test/JsonToTypes.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      cases: 100,
+      deaths: null,
+      hospitalized: 24,
+      ICU: 4,
+    }
+
+    const result: EmpiricalDatum = toEmpiricalDatum(input)
+
+    expect(result.deaths).toStrictEqual(0)
+  })
```
abrie: The `deaths`, `hospitalized` and `ICU` fields are sometimes null, and sometimes 0. The transformer should convert null value fields to 0.

src/algorithms/test/JsonToTypes.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+describe('JsonToTypes', () => {
+  // This test exists mainly for the linter to detect type errors.
+  it('caseCount.json can be parsed into EmpiricalData type', () => {
+    const testCountry = 'United States of America'
+    const caseCountsByCountry: CaseCountsByCountry = caseCountData
+    const caseCounts: CaseCounts[] = caseCountsByCountry[testCountry]
+    const result: EmpiricalData = toEmpiricalData(caseCounts)
+    expect(result).toHaveLength(caseCountData[testCountry].length)
+  })
```
abrie: This test demonstrates how to use the transformer. It mainly serves as a  way to detect type errors via the unit test task. The TypeScript compiler will show a warning during the test.

src/algorithms/test/JsonToTypes.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    const result: EmpiricalDatum = toEmpiricalDatum(input)
+    const expected = new Date('2020-03-17T00:00:00.000Z')
+
+    expect(result.time).toStrictEqual(expected)
+    expect(result.cases).toStrictEqual(100)
+    expect(result.deaths).toStrictEqual(0)
+    expect(result.hospitalized).toStrictEqual(24)
+    expect(result.ICU).toStrictEqual(4)
+  })
```
abrie: This test confirms that the date string is correctly turned into a Date.

src/algorithms/types/Param.types.ts
```diff
@@ -45,7 +45,7 @@ export interface AllParams {
 
 export type AllParamsFlat = PopulationData & EpidemiologicalData & SimulationData
 
-interface EmpiricalDatum {
+export interface EmpiricalDatum {
```
abrie: This needs to be exported so JsonToTypes can use it during transformation.

src/assets/data/CaseCounts.types.ts
```diff
@@ -0,0 +1,9 @@
+export interface CaseCounts {
+  time: string
+  cases: number | null
+  deaths: number | null
+  hospitalized: number | null
+  ICU: number | null
+}
+
+export type CaseCountsByCountry = Record<string, CaseCounts[]>
```
abrie: CaseCountsByCountry describes the whole of `case_counts.json`

src/components/Main/Main.tsx
```diff
@@ -10,6 +10,8 @@ import { Col, Row } from 'reactstrap'
 import { SeverityTableRow } from './Scenario/SeverityTable'
 
 import { AllParams, EmpiricalData } from '../../algorithms/types/Param.types'
+import { toEmpiricalData } from '../../algorithms/types/JsonToTypes'
```
abrie: Import the transformer...

src/components/Main/Main.tsx
```diff
@@ -63,7 +65,7 @@ async function runSimulation(
   }
 
   const ageDistribution = countryAgeDistributionData[params.population.country]
-  const caseCounts: EmpiricalData = countryCaseCountData[params.population.cases] || []
+  const caseCounts: EmpiricalData = toEmpiricalData(countryCaseCountData[params.population.cases] || [])
```
abrie: Use the transformer.
...And now the type error is gone.

src/algorithms/test/JsonToTypes.test.ts
ivan-aksamentov: @abrie I am not sure that this is correct. We need to check our science super-heroes @nnoll @rneher 

This json comes from a data repo (meaning, from external sources) and definitely requires validation and sanitization, but I think we may want to handle missing data differently than just assigning a neutral value. For example, consider a real-world scenario: unknown number deaths is not the same as 0 deaths. It's a huge difference in the context of epidemic.

Alternatively, we may need to validate here and crash early to catch data errors. Again, different fields have different meaning and value (as in some are more important than others), so they may need different handling.

src/algorithms/types/JsonToTypes.ts
```diff
@@ -0,0 +1,18 @@
...
+  const toNumber = (val: number | null) => (val === null ? 0 : val)
+
+  return {
+    cases: toNumber(input.cases),
+    hospitalized: toNumber(input.hospitalized),
+    deaths: toNumber(input.deaths),
+    ICU: toNumber(input.ICU),
+    time: new Date(input.time),
+  }
+}
```
ivan-aksamentov: @abrie If we are going for runtime validation here, I would prefer using a library wherever possible. Just to make sure this `toNumber()` don't grow hairs in the form of `toPositibeNumber()` and friends, to avoid writing the entire validation library from scratch.

Additionally, I would prefer declarative validation solutions. We currently use `yup` for Form validaton:
https://github.com/neherlab/covid19_scenarios/blob/73d3b94a578e39e2ea9ee01c5448a3e9a2bf0874/src/components/Main/validation/schema.ts#L12-L54
 and URL param folks are going to add `io-ts` (https://github.com/neherlab/covid19_scenarios/pull/53.)

This is probably an overkill for the `toEmpiricalDatum()` right now, but as we going forward, we need to validate all the external data sources (Form, LocalStorage https://github.com/neherlab/covid19_scenarios/pull/214, URL params https://github.com/neherlab/covid19_scenarios/pull/53, file import https://github.com/neherlab/covid19_scenarios/issues/11, more data files, network, etc.) in a cohesive manner, and to share validation schemas as much as possible to avoid updating multiple sources of truth.

src/algorithms/test/JsonToTypes.test.ts
abrie: Thanks for the close look. EmpiricalData type declares the deaths, hospitalized and ICU to be numbers; and the DeterministicLinePlot (which uses these values) checks only for 'truthiness', and never explicitly for a null. So converting to zero seemed like a reasonable assumption. But clearly there is a lot happening which cannot be inferred from the code – thank you for your explanation!

noleti: My understanding is also that we use null in the JSON if no data is available. That being said, this should ideally only happen at the beginning of a data series until some data is available for that column (as the columns are cumulative). I don't think we check for this when preparing the JSON so far. Is it possible to start plotting certain values only from a later day onwards, where values are not null any more?

ivan-aksamentov: @abrie Yep, so this is one of the cases where the type itself is incorrect, therefore validation could not be based on it.
Feel free to also check our data repo (in the submodule):
https://github.com/neherlab/covid19_scenarios_data
Your contributions have been very valuable here, so I am sure you could find something to hack there too :)

@noleti Yes, the missing values are currently not rendered, and they are not shown in the tooltip. This however would change if we would convert `null` to `0`.
In this particular case there is nothing wrong with `null` and I believe that data folks should send us `null` as it is. We just need to define what is the meaning of `null` and be prepared and handle it properly, both, in compile-time in typescript types and on runtime validation:
 - `null` but should be always preset? => data devs mistake  => crash early, before released
 - `null` and can be missing => OK =>  handle missing value during rendering

abrie:
Seems like this raises more issues than it resolves.

A better solution will evolve in the future, as the project standardizes on type schema/validation library: https://github.com/neherlab/covid19_scenarios/pull/243#discussion_r399019883

Ok to close this PR?

ivan-aksamentov:
@abrie Yes, I will close. But thanks for your efforts!

Right now I am thinking about schema versioning and how to validate. We may end up with something like JSON schema and code generation, to share across languages (TS, Python and the upcoming C++ WASM module). Feel free to take over it and play with it if you have time resources!

abrie:
@ivan-aksamentov sounds interesting – is there a branch or ongoing discussion somewhere?

ivan-aksamentov:
@abrie , nope, not yet. Only `yup` and `io-ts` was discussed so far, within the context of the app. But we need to think about data repo too.

PR #259, opened 2020-03-28
Added localstorage hooks for autorun, log scales, and show humanized results settings

### LINKED ISSUE #215: User UI toggleable settings are not persisted
# 🐛 Bug Report
User toggles such as autorun, log scales, and show humanized results are not persisted across page navigation/refresh.

## How to reproduce
Change a setting, refresh the page, observe that previous settings are not restored.

## 😯 Current Behavior
Settings aren't persisted across page reloads.

## 🤔 Expected Behavior
Settings should persist.

## 💁 Possible Solution
Persist UI configurable settings to local storage and restore on page load.

## Related
#214

ashwinGokhale:
## Related issues and PRs

 - Fixes #215 
- Relates to #214 

## Description
To make the following user settings persistent using the browser's local storage:
* Toggle autorun
* Log scales
* Show humanized results

* Main.tsx
* ResultsCard.tsx

## Testing

Open the site, toggle a UI setting, such as auto run, log graphs and show humanize. Refresh the page, observe that your UI preference is restored.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,7 @@
     "typescript-fsa-reducers": "1.2.1",
     "url-join": "4.0.1",
     "use-debounce": "3.4.0",
+    "use-persisted-state": "0.3.0",
     "yup": "0.28.3"
   },
   "devDependencies": {
@@ -169,6 +170,7 @@
     "@types/terser-webpack-plugin": "2.2.0",
     "@types/testing-library__cypress": "5.0.3",
     "@types/url-join": "4.0.0",
+    "@types/use-persisted-state": "0.3.0",
     "@types/webpack": "4.41.7",
     "@types/webpack-bundle-analyzer": "2.13.3",
     "@types/webpackbar": "2.6.0",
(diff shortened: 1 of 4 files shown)


ivan-aksamentov:
@ashwinGokhale Hi Ashwin! Thanks for the effort.

I don't quite like the idea of adding a hook for every option for several reasons:
 - low information to noise ratio, in terms of code usefulness 
 - hard to validate the correctness of the saved fields

I would prefer a more centralized solution with proper validation.
We currently have multiple ways to input data into he app: files (from data repo), forms, local storage, URL params, file uploads, possibly network API later on, and they often share the data schemas. So I thought it makes sense to adopt a unified validation mechanism. We currently use `yup` for forms, and `ts-io` incoming in URL land. Also we consider JSON Schemas for interation between data repo (Python) and the app.

Hook-per-field goes right against this paradigm, so I am not convinced we need this.
I will close for now, but fee free to submit a more centralized solution if you have time and resources.

PR #261, opened 2020-03-28
#101: stylelint --fix

patrikvarga:
## Description

`stylelint --fix` applied.

## Related issues

Contributes to https://github.com/neherlab/covid19_scenarios/issues/101

None.

## Testing

Build, test, run & play around.

### DIFF
--- a/src/components/Form/FormHelpButton.scss
+++ b/src/components/Form/FormHelpButton.scss
@@ -2,13 +2,13 @@ $radius: 16px;
 $color: #8b9dbe;
 
 .help-button {
-  background-color: $color;
-  border: none;
   width: $radius;
   height: $radius;
   padding: 0;
-  border-radius: $radius / 2;
   line-height: 0.5; // adjust this if you change the radius
+  background-color: $color;
+  border: none;
+  border-radius: $radius / 2;
 }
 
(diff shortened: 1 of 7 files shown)


ivan-aksamentov:
@patrikvarga Thanks Patrick! I need your opinion on stylelint config.
Did you find the ordering rules or any other rules unnecessary?

I think that once we enable linting during build (`yarn dev`) warnings should make developers to fix them. And ordering rules reduce diff sizes in long run. But in short run they are super-annoying.

Do we need to keep them? What do you think?

patrikvarga:
@ivan-aksamentov Thanks for asking. Here I don't have strong opinions either, but generally any rules/conventions will help consistency and easier understanding (in terms of knowing what to expect where) -- this one being no exception.

Such pains (especially linting/formatting/refactoring and similar bigger ones) are always annoying in the short term, and they are always even much more painful when not addressed early. :)

I guess my approach is the same as in the other issue: go for stricter rules, surface pains early, unless it causes too much friction; in which case it can easily be reverted. No decision is final, and I'm always happy to help out.

ivan-aksamentov:
@patrikvarga Thanks so much Patrik! You help has been invaluable.

Issue #148, opened 2020-03-23
Prettify About, Team and upcoming FAQ pages


ivan-aksamentov:
We need to make these a bit prettier, because right now it's a pure horror.

Unleash your imagination! 👨‍🎨

Related:
https://github.com/neherlab/covid19_scenarios/issues/147

P.S. FAQ page can be found on branch `feat/faq`

kimuraz:
Hi, can I work on this issue? * I'm not a designer*.  I've been working with frontend development for a while and saw some opportunities to adopt patterns that are commonly used on some other applications. If approved, I'd be glad to implement it.

Two points that concerns the whole application, would be:
1) Add a padding-top for all the containers with `20px`, making the spacing between the navbar and the content better.
2) Change a little bit the layout for the navbar, making your logo more visible and pushing the links to the right.
![image](https://user-images.githubusercontent.com/9062864/77799908-0fb20c00-7076-11ea-97c5-4298b132b6b9.png)

3) For the Team page, following the layout of the about page, something like:
![image](https://user-images.githubusercontent.com/9062864/77800613-7683f500-7077-11ea-941b-f012bee5a0f9.png)

4) For the FAQ page, just use an "Accordion" like [this](https://react-bootstrap.netlify.com/components/accordion/), that can be achievement on reactstrap as well. This can highlight the question and make it easier to the user to search for it, also focus on a single answer at the time.

What do you think?

ivan-aksamentov:
@kimuraz , sure, go ahead!
Make sure you roughly align the style with https://github.com/neherlab/covid19_scenarios/pull/195

(preview: https://covid19-scenarios-kr9r4tzcm.now.sh/)

You may save some time and take the team snippet from 
https://github.com/neherlab/covid19_scenarios/blob/master/README.md#initial-development
feel free to use MDX for pages or just components as well.

You can also add "Contributors section" to the team page. We auto-generate this snippet:
https://github.com/neherlab/covid19_scenarios/blob/master/src/assets/text/contributors-generated.md
with 
https://github.com/all-contributors/all-contributors

Important: all the main content in About and FAQ pages should be  in MD or MDX, so that Richard could update it more easily, without struggling with JSX :) Page layout can be in JSX.

And feel free to add yourself there to test how it works in the end :)
See their CLI documentation for how to do that.

kimuraz:
On it @ivan-aksamentov, will post some updates here when need. Thanks for the guidance.

[...]

PR #263, opened 2020-03-28
Layout improvement faq team

kimuraz:
## Related issues and PRs

Contributes to #148

## Description
This PR includes several commits changing the team (now acknowledgements) and FAQ pages.

- Navbar
- Team
- FAQ
- routes
- links

## Testing
UI checking


kimuraz:
> @kimuraz Thanks!
> 
> Few notes:
> 
> * [ ]  You don't need `react-markdown` package and to load and parse .md file on runtime. First of all it can be done on build-time. Second, loading files from GitHub and relying on their content is not going to work for us. What if someone change the README.md? The website would not render the page correctly.
> * [ ]  Gladly, there is a better approach. As I mentioned in the issue, we are using mdx.js and you can just import the .md file as a component. This way it becomes a part of the website bundle, just like any other React component. See for example `src/pages/About.mdx` (this is .mdx file, but it does not matter, you can import into .tsx files too).
> * [ ]  If you need to customize the imported component, you can pass custom components to override the defaults. See mdx.js docs ["Working with components"](https://mdxjs.com/getting-started#working-with-components)
> * [ ]  If needed, you can make a copy of the "Initial development" section of the REAMDEmd in a separate file, for example `src/assets/text/initial-development.md` and then import it. It will still be a duplicate, but at least when README.md is updated I can just copy-paste the update into `initial-development.md` and rebuild.

Thanks for the feedaback, I didn't realise the usage of the library on this way, but will replace it. My first though was to get the content from README during runtime because then it can be maintained in a single place, but of course the mdx will also fix it.

ivan-aksamentov:
@kimuraz  I understand your point about single place. duplicating is bad. However this content is unlikely to change, so it's okay.

ivan-aksamentov:
@kimuraz This is looking really good! Styling is a bit off here and there, but this is expected with markdown. It can be fixed later. Let me know when you are happy with the result and we can merge that.

kimuraz:
@ivan-aksamentov sure, thanks for reviewing, I'll do some more work on the contributors, I think it can be improved before, then I'll check lints and tests. I'll do it today evening.

kimuraz:
@ivan-aksamentov There are some issues on the CI but I saw there's a PR fixing them, once if it's merged before this one, I can rebase it.

kimuraz:
Rebased and passing the checks 👍

PR #266, opened 2020-03-28
Chore/update model figure

rneher:
## Description
We made a number of changes to the model that aren't reflected in the about page. This PR addresses some of this. 

Only about page.

## Testing
Spell check and correctness of equations.


rneher:
@nnoll Could you check this?

PR #267, opened 2020-03-28
Github Action: Update submodule and jsons

tryggvigy:
## Related issues and PRs

 - https://github.com/neherlab/covid19_scenarios/discussions/268

## Description
Add workflow to update submodules every 1h. This will keep `covid19_scenarios_data` in up to date.

This PR is not final but demonstrates how this could be done. To get all the way there we would need

**Checklist**
- [ ] Introduce `requirements.txt` file containing the dependencies of `covid19_scenarios_data` (`requests`, `numpy`, ...)
- [ ] Install python3 in action
- [ ] Install pip in action
- [ ] Run `pip install -r requirements.txt` in action
- [ ] Action should succeed.

None of which are difficult but should be discussed in https://github.com/neherlab/covid19_scenarios/discussions/268

### DIFF
--- a/.github/workflows/update-submodule-and-jsons.yml
+++ b/.github/workflows/update-submodule-and-jsons.yml
@@ -1,0 +1,29 @@
+name: 'Update Submodule and Jsons'
+on:
+  schedule:
+    - cron: '0 * * * *'
+
+jobs:
+  sync_submodules:
+    steps:
+      - name: Checkout submodule
+        uses: srt32/git-actions@v0.0.3
+        with:
+          args: git submodule update --init --recursive
+      - name: Update jsons
+        run: |
+          python covid19_scenarios_data/parse_all.py --fetch \
+          --output-cases path/case-counts.json \
+          --output-population path/population.json
+      - name: Commit files
+        run: |
+          git config --local user.email "<authors email here>"
+          git config --local user.name "<authors name here> Action"
+          git add .
+          git commit -m "Update submodule and jsons"
+        continue-on-error: true
+      - name: Push changes
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+        continue-on-error: true


nnoll:
This would help me tremendously as I'm very bad at updating submodules...

rneher:
This seems like great idea. we have been juggling how to best handle updates to the data. This turns out to be a little tricky since the different sources we pull from are not extremely stable and changes in these sources tend to break things once in a while. but we definitely have to come up with a more stable solution. Tagging @noleti. Would be great to come up with a better solution here.

tryggvigy:
The action can run every 1h

If
```
python covid19_scenarios_data/parse_all.py --fetch \
          --output-cases ./src/assets/data/case-counts.json \
          --output-population ./src/assets/data/population.json
```

is prone to be flaky we have at least 24 chances to run in successfully each day.

rneher:
One solution would be to generate `json`s from the tsv independent of `fetch`. `tsv` would be committed when updating the data repo after spot-checking. This would separate the fetching from external sources from the update of the app. 

If this seems like a good idea, I'll make an issue in `covid19_scenarios_data`.

tryggvigy:
@rneher That sounds reasonable. Am I understanding right that you feel that keeping the human verification step before updating the app is important?

noleti:
Having all parsers generate .tsv in the Region/sub-region/country/ folders, and then reparsing them to generate the json is of course possible. I would prefer to fix issues in the .json generation (just fixed some today) and skip the dumping to .tsv, but of course I can understand if you prefer to do manual verification first. One open issue would be the cds.py parser, which generates world-wide data (like ecdc.py). Before seing this issue, I had just removed the cds.tsv generation (because it requires some dirty hacks), but I can of course re-enable it. Should parsers like ecdc.py and cds.py dump into one world.tsv, or individual top-level .tsvs, or eventually generate the correct folder hierarchy?

rneher:
I made an issue here: https://github.com/neherlab/covid19_scenarios_data/issues/63

I like the tsv because `git diff case-counts/*` shows me exactly which data got added for which file. But I agree that this is a little backwards.

noleti:
for diffing .json: you can actually tell git to use a pretty printer before diffing. The following should set this up for any repo you have

```
echo "*.json diff=json" >> ~/.gitattributes
git config --global core.attributesfile ~/.gitattributes
git config --global diff.json.textconv "python -m json.tool"
```

ivan-aksamentov:
@tryggvigy @nnoll @rneher @noleti 
This touches a pinful point of mine.

Alright, actions are cool and let you do crazy stuff, but this particular one introduces a race condition in the build system, is a very bad idea and is a workaround for my silly mistake of introducing a submodule.

Consider the following scenario:
 - a build on CI server is triggered automatically
 - someone commits into the data repo a change that also requires a commit into the app repo, in order for app to function
 - a commit is required in the app repo haven't been committed yet (these two commits are not a single transaction)
 - the proposed Action updates the submodule
 - the CI server proceeds with cloning the repo, pulling the new data but the app that haven't been updated yet
 - build succeeds because there are no any checks on data correctness whatsoever
 - production crashes, requiring manual intervention and messing with submodules in the middle of night

There is probably a dozen more of scenarios like that. Concurrency is hard.

Instead, until it's too late, we should:
 - [ ] move the data back into the app repo https://github.com/neherlab/covid19_scenarios/issues/282
 - [ ] introduce a schema to validate data flowing between Python and Typescript codebases (plus the upcoming C/Wasm thingy?) as well as between different ways of data inputs https://github.com/neherlab/covid19_scenarios/issues/283

Am not inclined to merge this.
Comments? Objections?

rneher:
We have moved the data back into app. schemas are being worked on and a draft PR is up:
https://github.com/neherlab/covid19_scenarios/pull/355

hongaar:
@nnoll not sure if re-opening my PR (https://github.com/neherlab/covid19_scenarios_data/pull/47) in this repo would still be useful after reading up in the discussion above?

ivan-aksamentov:
@hongaar This might be useful after we figure out what has to be run on updates and how to automate this.

In particular we have no validation or any sanity checks for the generated data whatsoever. On every update, Richard and Nick basically have to eyeball the data. So autoupdates may be dangerous right now. 

We will eventually need some sort of solution, because the current practice is not sustainable. 
Thoughts and ideas are welcome. Feel free to open a thread on Discussions tab and/or an issue.

PR #271, opened 2020-03-28
Set an upper limit on eslint warnings so that more cannot be merged

gj262:
## Related issues and PRs

Contributes to #101

## Description

There are currently 114 eslint warnings. With this change future PRs will not be able to increase that count. The goal would be to gradually reduce this high water number as lint errors are fixed until an eslint warning free codebase is reached 🎉 

Lintiness.

## Testing

This PR has an initial max count of 113 which should fail travis CI. Then I'll change it and it should pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_script:
 
 script:
   - yarn test:nowatch
-  - yarn eslint --max-warnings 113 || (echo ERROR New eslint errors were introduced by these changes && exit 1)
+  - yarn eslint --max-warnings 114 || (echo ERROR New eslint errors were introduced by these changes && exit 1)
 
 before_deploy:
   - yarn prod
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ before_script:
 
 script:
   - yarn test:nowatch
+  - yarn eslint --max-warnings 113
 
 before_deploy:
   - yarn prod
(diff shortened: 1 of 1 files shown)


gj262:
@patrikvarga you might appreciate this.

ivan-aksamentov:
@gj262 @patrikvarga This is 100% correct move for most web apps. But here I am a bit hesitant, because we have hardcore scientists working on the project and they may not know much about linting.

I would prefer to cleanup after them, rather than delay them in iterating on the algorithm.
But I guess we can always jump on their branches, cleanup and then merge, right? That should work, I think.

gj262:
> @gj262 @patrikvarga This is 100% correct move for most web apps. But here I am a bit hesitant, because we have hardcore scientists working on the project and they may not know much about linting.
> 
> I would prefer to cleanup after them, rather than delay them in iterating on the algorithm.
> But I guess we can always jump on their branches, cleanup and then merge, right? That should work, I think.

Yeah I hear what you are saying re trade offs. I guess we will see if it works now that you merged it ;) If the scientists don't get the help they need to move fast we can easily back it out.

patrikvarga:
@gj262 Thanks for the heads-up.

@ivan-aksamentov I don't really have strong opinions here. Let's see what works; I'm happy to do the cleanups either way, especially if that makes it easier for others to contribute.

PR #272, opened 2020-03-28
Refactor local storage

maktarsis:
## Related issues and PRs
#215 
 - 

## Description
I really curious about consistency and I want to align with maintainer. If this PR is ok, I can go further and implement StorageService which is an abstraction and can be used upon any type of storage.

Storage

1) if something is wrong, I can easily fix it
2) I made a small PR to understand the policy, how interesting it is for you to have consistency
3) checkout description section, I will be pleased to implement storage service
4) In addition, I refactored some things in my OL that could lead to errors in the future.

### DIFF
--- a/src/components/LandingPage/LandingPage.tsx
+++ b/src/components/LandingPage/LandingPage.tsx
@@ -1,19 +1,30 @@
-import React, { useState } from 'react'
-
+import React, { useState, useCallback } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import { Link } from 'react-router-dom'
-import './LandingPage.scss'
-import promoImage from './tool.jpg'
+
+import LocalStorage, { LOCAL_STORAGE_KEYS } from '../../helpers/localStorage'
 import { setShouldSkipLandingPage } from '../../state/ui/ui.actions'
-import { SKIP_LANDING_PAGE_KEY } from '../../state/ui/ui.reducer'
 import { State } from '../../state/reducer'
+import promoImage from './tool.jpg'
+import './LandingPage.scss'
 
 function LandingPage() {
   const [isSkipLandingChecked, setSkipLandingChecked] = useState(false)
   const showSkipCheckbox = useSelector(({ ui }: State) => !ui.skipLandingPage)
   const dispatch = useDispatch()
+  const onLinkClick = useCallback(() => {
+    dispatch(setShouldSkipLandingPage({ shouldSkip: true }))
+
+    if (isSkipLandingChecked) {
+      LocalStorage.set(LOCAL_STORAGE_KEYS.SKIP_LANDING_PAGE, String(true))
+    }
+  }, [dispatch, isSkipLandingChecked])
+
+  const onCheckboxChange = useCallback((e) => {
+    setSkipLandingChecked(e.target.checked)
+  }, [])
+
   return (
-    <>
       <div className="landing-page">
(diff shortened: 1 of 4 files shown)


ivan-aksamentov:
@rcbevans Could you please have a glance and merge it it's okay (I've sent you write access some time ago)

rcbevans:
> @rcbevans Could you please have a glance and merge it it's okay (I've sent you write access some time ago)

Sure, I'll take a look.

src/helpers/localStorage.ts
```diff
@@ -1,24 +1,19 @@
...
   AUTORUN_SIMULATION: 'autorun-simulation',
   LANG: 'lang',
   LOG_SCALE: 'log-scale',
   SHOW_HUMANIZED_RESULTS: 'show-humanized-results',
   SUPPRESS_DISCLAIMER: 'suppress-disclaimer',
-}
+  SKIP_LANDING_PAGE: 'skip-landing-page'
+} as const
+
+export type LocalStorageKey = typeof LOCAL_STORAGE_KEYS[keyof typeof LOCAL_STORAGE_KEYS]
```
rcbevans: Nice 👍

src/components/LandingPage/LandingPage.tsx
```diff
@@ -1,32 +1,38 @@
...
 
 function LandingPage() {
   const [isSkipLandingChecked, setSkipLandingChecked] = useState(false)
   const showSkipCheckbox = useSelector(({ ui }: State) => !ui.skipLandingPage)
   const dispatch = useDispatch()
+  const onLinkClick = useCallback(() => {
+    dispatch(setShouldSkipLandingPage({ shouldSkip: true }))
+
+    if (isSkipLandingChecked) {
+      LocalStorage.set(LOCAL_STORAGE_KEYS.SKIP_LANDING_PAGE, String(true))
```
rcbevans: No need to cast to a string here.  This results in the LocalStorage value being `"true"`, rather than `true`.  If this value is changed to `"false"` (as opposed to `false`), the ui reducer cast to a boolean actually doesn't work as expected, because `Boolean('false')` is actually `true`.  The UI reducer as is here will actually return `true` for anything except `null`/`undefined`/`''` (the initial state), because any string that isn't empty is truthy.

I'm signing off because the behavior functions as expected, and requires someone to manually change the LocalStorage value for it to misbehave.

maktarsis: if we do `JSON.parse` of `getItem` then we will get `true` rather than `"true"`

we should parse a value from storage in any normal situation

rcbevans: Stringification of boolean `true` and string `'true'` produce different values, and so the `JSON.parse(stringified)` produces a different recovered value.

```
JSON.stringify(true)
"true"
JSON.stringify('true')
""true""
```

Try it out in your console.  In your case, because you stringified a string, you get back a string, which will always be truthy when cast to a bool.

maktarsis: yeah, JSON.stringify works as you said
but it doesn't work as `String` constructor does
https://prnt.sc/roq6rm

but now I understand your remark
at some point I have refactored `localStorage.setItem` for `LocalStorage.set` which does `JSON.stringify` under the hood. 
Yeah, I'll fix this in my next PR
thanks

PR #274, opened 2020-03-28
Fix regression 'interpolation within the timerange occurs as expected.'

abrie:
## Related issues and PRs

This regression was previously fixed in #238 

## Description

The TimeSeries.test.ts `'interpolation within the timerange occurs as expected.'` test fails sometimes.

image

The test uses toMatchSnapshot(), which compares the results using string representations.
The test is fixed by instead comparing the arrays using toBeCloseTo() assertions.

TimeSeries.test.ts unit tests.

## Testing

`yarn test:lint TimeSeries`

### DIFF
--- a/src/algorithms/utils/TimeSeries.test.ts
+++ b/src/algorithms/utils/TimeSeries.test.ts
@@ -91,7 +91,28 @@ describe('TimeSeries', () => {
         return interpolator(date)
       })
 
-      expect(result).toBeArrayOfSize(intervalCount).toMatchSnapshot()
+      expect(result).toBeArrayOfSize(intervalCount)
+
+      /* toMatchSnapshot() doesn't work well with floating point numbers.
+       * So we need to manually iterate and assert using toBeCloseTo().
+       */
+
+      const expected = [
+        1,
+        2.6354594390679185,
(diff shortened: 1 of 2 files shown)


abrie:
@ivan-aksamentov Absolutely agree. Needing to manually modify tests is a good way to discourage their use.

My first thought was a custom snapshot matcher to parse the snapshot string into an array using `JSON.parse()`. Then we could employ the looping `isCloseTo()` assertion. But I've found no way to access the snapshot string. A custom Jest matcher only enables you to modify the expected result before passing it to the base implementation of toMatchSnapshot() – there is no way to override the comparison logic in a meaningful way, nor access the snapshot contents. Inspecting the [Jest snapshot code](https://github.com/facebook/jest/blob/2a4b073ff7a607bda4380cb0a544a8521719bc4a/packages/jest-snapshot/src/index.ts#L280) hasn't revealed any obvious hack.

abrie:
@ivan-aksamentov I spoke too quickly. There is indeed a way to access the snapshot contents.

Tagging #281 here to show that work is in progress.

ivan-aksamentov:
Closing this now that we have #295 merged

PR #277, opened 2020-03-28
#101/Add "arrowParens": "always" to prettier config.

abrie:
## Related issues and PRs
#101 , #260 

## Description

The linter is configured to want parens around arrow function parameters, but Prettier was not configured to add them. Consequently there are regressions like #260. This PR fixes that.

Linting

## Testing

Edit a source file. Try to create an arrow function without parenthesized parameters. Your editor should automatically add the parentheses.

### DIFF
--- a/config/.prettierrc
+++ b/config/.prettierrc
@@ -4,5 +4,6 @@
   "quoteProps": "consistent",
   "semi": false,
   "singleQuote": true,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "arrowParens": "always"
 }


ivan-aksamentov:
@abrie 
We've recently updated Prettier to 2.0 to support some of the neat Typescript 3.8 features.
Prettier 2.0 now has this setting default:
https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev

The problem is that some devs and their editors may still use Prettier 1.0, either because they don't `yarn install` often enough, or if they have it installed globally. Not sure how to enforce that. Upgrading was probably a bad idea.

abrie:
@ivan-aksamentov 
Ah! Understood. I think this is probably a sign of the project reaching some sort of critical size. A phase change, if you will.

> We've recently updated Prettier to 2.0 to support some of the neat Typescript 3.8 features.
> Prettier 2.0 now has this setting default:
> https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev
> 
> The problem is that some devs and their editors may still use Prettier 1.0, either because they don't `yarn install` often enough, or if they have it installed globally. Not sure how to enforce that. Upgrading was probably a bad idea.

ivan-aksamentov:
@abrie Yeah, plus me not paying enough attention to PRs. Way too many for a single engineer. Feel free to hop in on reviews too :)

PR #286, opened 2020-03-29
moved persisting settings logic to redux store

ashwinGokhale:
## Related issues and PRs

 - Related to #272 

## Description

Moved the global settings to a redux reducer because settings are a global state and moving them to the redux store makes sure that redux is the single source of truth for getting and setting these values. I think that manually updating settings in localstorage directly could lead to some problems down the road, (like someone forgetting to update it) so I moved the handling of persisting the settings to localstorage as a side effect of updating the redux store.

Main.tsx
ResultsCard.tsx
src/state

## Testing

1) Open the site, Refresh the page, observe that your UI preference is restored.
1.1) Checks for keys not in localstorage
2) Open the site, toggle a UI setting, such as auto run, log graphs and show humanize. Refresh the page, observe that your UI preference is restored. 
2.1) Checks for keys in localstorage and values are updated


src/state/settings/settings.reducer.ts
```diff
@@ -0,0 +1,40 @@
...
+  logScale: boolean
+  showHumanized: boolean
+}
+
+const AUTORUN_SIMULTAION_DEFAULT = false
+const LOG_SCALE_DEFAULT = true
+const SHOW_HUMANIZED_DEFAULT = true
+
+export const settingsDefaultState: SettingsState = {
+  autorunSimulation: Boolean(LocalStorage.get(LOCAL_STORAGE_KEYS.AUTORUN_SIMULATION ?? AUTORUN_SIMULTAION_DEFAULT)),
```
rcbevans: Close paren is in the wrong place here, should be before the null coalesce.

src/state/settings/settings.actions.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface LogScaleSettingsParams {
+  logScale: boolean
+}
+
+export interface ShowHumanizedSettingsParams {
+  showHumanized: boolean
+}
+
+export const setAutorunSimulation = action<AutorunSettingsParams>('SETTINGS_SET_AUTORUN_SIMULATION')
```
rcbevans: There are a few other settings that are currently persisted to LocalStorage, `SUPPRESS_DISCLAIMER` and `SKIP_LANDING_PAGE`.  You may want to include these for consistency.

ashwinGokhale: > There are a few other settings that are currently persisted to LocalStorage, `SUPPRESS_DISCLAIMER` and `SKIP_LANDING_PAGE`. You may want to include these for consistency.

SKIP_LANDING_PAGE is part of the ui settings reducer. Do you think it would be a good idea to combine SKIP_LANDING_PAGE and SUPPRESS_DISCLAIMER into a single reducer. If so, should I combine them into the ui reducer or the settings reducer. I would prefer settings reducer IMO.

ivan-aksamentov:
HI @ashwinGokhale , this sounds reasonable, however right now I don't want to do that.

We've got some important long-running features baking right now and I don't want to create any additional rebasing noise for these folks, at least not at a cost of "conceptual" improvements. Remember, we are an outbreak planing tool running in the middle of a pandemic, so our priorities might be a bit weird.

Centralizing local storage is a must, because we are going to centrally validate it as well, and then maybe gather all the options in a "Settings" UI, but we will do this a bit later. Maybe in a week or so, depending on how major features go.

I will mark this PR in the validation issue to not forget.

Issue #283, opened 2020-03-29
Introduce a unified type schema


ivan-aksamentov:
# 🙋 Feature Request

We need to have a way to ensure on build-time that the changes introduced between data processing and the aligned. Right now the only way s to discover the issues on runtime (whether in development of later on in the deployed app).

## 🔦 Context

Currently Typescript side have a set of typings written manually. Python side has a set of matching classes written manually. All that is not connected or verified in any way. This is brittle, and error-prone. Updating these manually is a waste of time. There are tools that can do that automatically.

We have a variety of ways the data can enter the app (inputs, deserialization):
 - form input (currently validated with `yup` schema)
 - URL params (currently not validated, but there was a discussion to introduce `io-ts` https://github.com/neherlab/covid19_scenarios/pull/338 and `jsurl` https://github.com/neherlab/covid19_scenarios/pull/329)
 - upcoming import from file(s) https://github.com/neherlab/covid19_scenarios/issues/11 https://github.com/neherlab/covid19_scenarios/pull/381
 - local storage - params and user settings (currently not validated)
 - possibly network API soon 

As you see we are running our app on pure luck right now, hoping that user wouldn't feed us bad data. That has to be changed.

Same for outputs (outputs, serialization):
 - sharing a URL with params (a reverse of URL params inputs, above)
 - file export (similar, but not necessarily a strict reverse of file import)
 - local storage (a reverse of local storage inputs)
 - possibly network API soon 

## 😯 Describe the feature

 - [x] have a set of schemas, that declaratively describe the various types in the app
 - [ ] commit these schemas into version control and thoroughly review every change
 - [ ] automatically generate Typescript typing and Python classes from these schemas on every build (remove corresp. manually written  types)
 - [ ] validate inputs and outputs against the same schema on runtime
 - [ ] automatically generate documentation from schemas (there are tools for that)m  and publish this documentation
 - [ ] final serialization objects should contain version information, so that we could in the new versions of the app handle inputs saved from earlier versions

## 💁 Possible Solution

 - [ ] [JSON schema](https://json-schema.org/) is a typical choice here
 - [ ] [Swagger/OpenAPI](https://swagger.io/docs/specification/about/) for describing network APIs (superset of subset of JSON schema)

## Related

 - https://github.com/neherlab/covid19_scenarios/pull/267
 - https://github.com/neherlab/covid19_scenarios/issues/282

ivan-aksamentov:
[...]

Issue #222, opened 2020-03-26
Cloud based syncing/backup of scenarios/user preferences


rcbevans:
# 🙋 Feature Request
Add a mechanism of backing up/restoring using preferences and configured scenarios to the cloud for use across multiple clients.

## 🔦 Context
Roam my preferences and custom scenarios with me.

## 😯 Describe the feature
Add login/logout which will synchronize the values in LocalStorage to a cloud based data store.  On login, these cloud values will be pulled, replacing/merging with current configuration.

## 💁 Possible Solution
Firebase has an all in one auth/signup/social media sign in, and a very very easy to use json based realtime database which can be used in a very similar manner to LocalStorage, and can even fire events on cloud database updates, making realtime sync across devices seamless.

miksa1987:
@rcbevans I could try implementing this? I think this could be done with Firebase auth and Firestore, as I find Firestore even easier to use than Realtime database.

Some points I'm thinking:
- There could be a users collection, where a document is created when user is created, with the document id being the user's uid.
- On logging in, the user's data is fetched from the document, which is found by the user's uid.
- The values are put into localStorage, and whenever some of those values are updated, they will be also updated to Firestore.

At first I could just do an email signin, we could check out other signin options later.

rcbevans:
We might need someone at neherlab to create and maintain ownership of the firebase account.  @ivan-aksamentov?

Some things to think about:

The login/logout, and fresh start of data experience.
We'll need some kind of privacy policy, and GDPR compliance, allowing users to obtain/delete their data freely.

We want to be very careful around user credentials, if doing email/password based login (I always prefer oauth based login for this reason, so I never have to touch a password).

rcbevans:
Also think about the multi-scenario, and consider that when designing the data structure.

Most important is ensuring that the firestore has role based access controls so users can only read/write their own user data.

miksa1987:
I could experiment with my own Firebase account, it could be probably easily switched to neherlab's by using environment variables. Of course, neherlab account should have auth and firestore ready when starting to use that.

Regarding user experience, I was thinking about a login/logout button somewhere on top of page, that would trigger a popup form for login(like the disclaimer when arriving to the site). The account creation form would be something similar.

I'm wondering if there should be some kind of indicator for being logged in, where the user could go to check/delete their data? Maybe an additional component for the top bar which could house the login/logout button as well?

[...]

Issue #181, opened 2020-03-24
Multiple persistent shareable scenarios


gj262:
# 🙋 Feature Request

The current implementation shows a single scenario and allows the user to change it and view the results. The scenario input can be shared through the app URL.

This feature allows for multiple persistent and shareable scenarios as the title suggests :).

## 🔦 Context

This feature evolves from issues #11 and #31. And while not implementing these precisely it makes progress in that direction and evolves to meet the items in some of comments in those issues. Hence a new feature!

## 😯 Describe the feature

There will be tabs below the main header, one to customize scenarios and one for each saved scenario. 

## 💁 Possible Solution

Initially users will see no tabs <sup>1</sup>. There will be a **save action** that is clickable once inputs are modified. Clicking save will present a modal requesting a name and owner email for the scenario to be saved. Once saved, the scenario (name, owner, input, result) will persist to local storage. And the view will change to tabs <sup>1</sup>. The in view tab is titled "Customize" with the prior scenario state and a new tab will appear for the saved scenario titled with that scenarios name . 

The view of the saved scenario will include a **delete action** protected with an are you sure modal.

There will be a **share action or link** in each tab which provides the saved scenario (name, owner, input, result) as a URL.

If a user clicks on a shared URL the application state will show their own saved scenario tabs plus an extra tab with this shared scenario titles name + (owner). They can choose to save this shared scenario or close the tab or delete it.

As a first pass, saved scenarios should be immutable i.e. the inputs are not changeable and are disabled in the saved scenario view. Otherwise there is a possibility for confusion when scenarios are shared and then edited by various participants. Saved scenarios could be cloned and the clones input could be changed. <sup>2</sup>.

Initially the actions would follow the button format currently in use for Run and Export. However this should be improved with a separate issue.

URL state will match the in view tab scenario.

## Future 

There is now a local store of named and owned scenarios. This could be stored remotely. There could be a page to manage these scenarios. There could be views to compare the results.

## To Be Resolved

- [ ] Tabs will not work for mobile
- [ ] Can a result be URL serialized?

## Notes

1. Having tabs appear for the first saved scenario is lame UX. But the alternative of always having a single "Customize" tab is also somewhat lame. Will need to experiment.
2. Follow some existing art for best practices.

rcbevans:
[...]

PR #293, opened 2020-03-29
Multiple persistent scenarios

gj262:
## Related issues and PRs

Relates to #11 #181 #222.

## Description

This change is an initial step towards multiple scenario support. Customized scenarios can be saved and then appear in individual tabs. The scenario view is still managed by a single `<Main />` instance. This allows a switch between scenario tabs to show the result without jarring re-renders. `<Main />` handles this by segmenting the state/reducer by scenario. I would have liked to refactor that a bit more but resisted that urge in order to minimize change.

From this first step one could: Have another page to compare and manage saved scenarios along the lines of the feature requested in #11.

### To do

- [x] Include the severity table as saved input parameters
- [x] Delete
- [x] Sharing
- [ ] Edit
  - [ ] Rename
  - [ ] Revert  
- [ ] Delete Y/N protection
- [ ] Mobile view (use a menu instead of tabs?)
- [ ] Tiny URLs
- [ ] Mine -v- Shared
- [x] Fix types. My typescript skillz cannot handle dealing with dynamic action payloads.
- [ ] Use a type schema for (de)serialization #283.
- [ ] Add creation date to shared scenarios

Main scenario view.

## Testing

This is a WIP but at present it appears that the individual scenario tabs correctly represent the state of their parameters and produce the expected result.


gj262:
Please ignore this PR for now. I've managed to majorly mess up state.

gj262:
I came up with a cleaner and less broken way to have `<Main />` handle many scenarios. In the process I broke everything else. Fixes on the way.

src/components/Main/Main.scss
```diff
@@ -1,9 +1,10 @@
...
-.export-button,
-.run-button {
-  width: $button-width;
-  margin-left: $button-spacing;
+.controls button:not(:last-child) {
+  margin-right: $button-spacing;
+}
+
+.controls button {
+  min-width: $button-width;
```
gj262: Allows buttons to grow a little for different languages to avoid wrap. See e.g. Deutsch.

src/components/Main/Main.tsx
```diff
@@ -139,27 +163,54 @@ function Main() {
...
+      setSeverity(incomingParams.severity)
+    }
+  }, [incomingParams, cancelSetScenarioToCustom, cancelDebouncedRun])
+
+  useEffect(() => {
+    if (incomingResult) {
+      setResult(incomingResult)
+    }
+  }, [incomingResult])
+
```
gj262: With these changes (more or less), the parent component <MultipleScenarios /> can control <Main />

src/components/Main/Results/AgeBarChart.tsx
```diff
@@ -103,10 +103,25 @@ export function AgeBarChart({ showHumanized, data, rates }: SimProps) {
...
+                  fill={colors.overflow}
+                  name={t('peak overflow')}
+                  isAnimationActive={false}
+                />
+                <Bar
+                  dataKey="totalFatalities"
+                  fill={colors.fatality}
+                  name={t('total deaths')}
+                  isAnimationActive={false}
+                />
```
gj262: Animation ruins performance when switching between scenarios.

src/components/MultipleScenarios/MultipleScenarios.tsx
```diff
@@ -0,0 +1,225 @@
...
+        scenarios: savedScenarios.scenarios.filter((scenario) => scenario.id !== toDelete),
+      })
+      setScenarios(scenarios.filter((scenario) => scenario.id !== toDelete))
+    }
+  }
+
+  function handleChangedParameters(params: ScenarioParams) {
+    // Note: isEqual handles Date() objects while lodash.isEqual does not.
+    const different = !isEqual(activeScenario.params, params)
+
```
gj262: Be very wary abut circular updates.

gj262:
Hi @ivan-aksamentov, when you get a chance can you take a look at this PR? In its current form it works but would not be mergeable due to performance and other issues. For instance, the graph updates when switching scenario tabs do not work too well when the axis change. I can fix that and any other issues but I don't want to spend any more time on it if this feature is not something the team wants to pursue.

rneher:
@gj262 lots of great ideas here! The share, save, and tab options are very useful.
But we currently have the problem that the query string isn't pass through the landing page. this affects everything that relies on query strings as far as I can tell.

dkozar:
> But we currently have the problem that the query string isn't pass through the landing page. this affects everything that relies on query strings as far as I can tell.

@rneher not necessarily. For instance, it doesn't affect running the scenario another new tab. The feature could be checked here: https://covid19-scenarios-cfau3dmod.now.sh/

(PR: https://github.com/neherlab/covid19_scenarios/pull/368)

The query string not surviving the landing page is a separate issue. I can take a look at it.

image

gj262:
> @gj262 lots of great ideas here! The share, save, and tab options are very useful.
> But we currently have the problem that the query string isn't pass through the landing page. this affects everything that relies on query strings as far as I can tell.

Yeah, I've noticed that bug. That'll need fixing before this would fly. This PR is probably too adventurous to lead to a merge but it has allowed me to wrap my head around the app state. So that is something :)

gj262:
Closing as out of date & not in scope.

Issue #281, opened 2020-03-29
Introduce custom Jest snapshot matchers


ivan-aksamentov:
# 🙋 Feature Request

Introduce a set of custom Jest snapshot matchers that could work for our assertions better.
https://jestjs.io/docs/en/expect#custom-snapshot-matchers
https://github.com/maasencioh/jest-matcher-deep-close-to

We use a lot of floating point numbers and they are not correctly compared by default matchers. 

It is preferred to not go too deep into the app-specific data structures ant to not hard-code object layouts etc., if possible.

It is desirable to parameterize the "closeness" of the compared values, either on absolute difference (commonly called "epsilon") or on Units in the Last Place (ULPs) in its IEEE-754 representation.

## 🔦 Context

We want to avoud hardcoding expectations and comparing them with `.toBeClose()`because it is very tedious to fix all the tests after every small change in the algorithm. 
We want to update snapshots with `jest -u` instead. However, this require a snapshot matcher that understands floating point numbers.

## Related

 - https://github.com/neherlab/covid19_scenarios/pull/274
 - https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html

ivan-aksamentov:
@abrie Could you please let me know if #327 is still useful? Could we revive and rebase it?

ivan-aksamentov:
Fixed in #327

PR #295, opened 2020-03-29
#274/WIP Implement a Jest extension to compare arrays of floats to snapshots.

abrie:
## Related issues and PRs

Addresses concerns raised in #274 by implementing #281.

## Description

Jest snapshots are very useful because they can be updated from the test runner UI. But the default Jest snapshot assertion compares serialized strings. That causes problems if floats need to be compared. This PR implements a custom snapshot matcher for arrays of float values. It works by serializing expected results as a JSON array into the snapshot file. The snapshot can be deserialized and checked using a loop of toBeCloseTo() assertions. With this extension algorithm developers can use the Jest UI to add/update/remove snapshot results without modifying test logic.

This is a hack – it touches internal, private, Jest state. No guarantee this will continue to work if Jest is upgraded.

Unit tests

## Testing

`yarn test TimeSeries`

### DIFF
--- a/src/algorithms/utils/TimeSeries.test.ts
+++ b/src/algorithms/utils/TimeSeries.test.ts
@@ -1,6 +1,106 @@
 import { uniformDatesBetween, makeTimeSeries, updateTimeSeries } from './TimeSeries'
 import { interpolateTimeSeries } from '../run'
 
+import * as fs from 'fs'
+const { utils } = require('jest-snapshot')
+
+expect.extend({
+  toBeCloseToArraySnapshot(received: number[]) {
+    const { testPath, currentTestName, snapshotState } = this
+    snapshotState._counters.set(currentTestName, (snapshotState._counters.get(currentTestName) || 0) + 1)
+    const count = Number(snapshotState._counters.get(currentTestName))
+    const key = utils.testNameToKey(currentTestName, count)
+    const expected = snapshotState._snapshotData[key]
+
+    /* If this isn't done, Jest reports the test as 'obsolete' and prompts for deletion. */
+    snapshotState._uncheckedKeys.delete(key)
+
+    /* If the snapshot isn't JSON, then return an empty array. Otherwise
+     * the test will fails in an inscrutable way.
+     */
+    const tryParse = (str: string) => {
+      try {
+        return JSON.parse(str)
+      } catch (e) {
+        return []
+      }
+    }
+
+    /* If the expected field is blank, then this is probably a new snapshot. */
+    const expectedDeserialized = expected ? tryParse(expected) : []
+
+    const pass =
+      expectedDeserialized.length > 0 /* must check this because every() returns true if empty array */
+        ? received.every((_, idx) => Math.abs(expectedDeserialized[idx] - received[idx]) < 10 ** -2 / 2)
+        : false
+    const hasSnapshot = expected !== undefined
+    const snapshotIsPersisted = fs.existsSync(snapshotState._snapshotPath)
+    const receivedSerialized = JSON.stringify(received, null, 2)
+
+    if (pass) {
+      /* Test passed, now update the snapshot state with the serialize snapshot.
+       * Technically this is only necessary if no snapshot was saved before. */
+      snapshotState._snapshotData[key] = receivedSerialized
+    }
+
+    /* This nested mess is derived the Jest snapshot matcher code:
+     * https://github.com/facebook/jest/blob/4a59daa8715bde6a1b085ff7f4140f3a337045aa/packages/jest-snapshot/src/State.ts
+     */
+    if (
+      (hasSnapshot && snapshotState._updateSnapshot === 'all') ||
+      ((!hasSnapshot || !snapshotIsPersisted) &&
+        (snapshotState._updateSnapshot === 'new' || snapshotState._updateSnapshot === 'all'))
+    ) {
+      if (snapshotState._updateSnapshot === 'all') {
+        if (!pass) {
+          if (hasSnapshot) {
+            snapshotState.updated++
+          } else {
+            snapshotState.added++
+          }
+          snapshotState._addSnapshot(key, receivedSerialized, { error: undefined, isInline: false })
+        } else {
+          snapshotState.matched++
+        }
+      } else {
+        snapshotState._addSnapshot(key, receivedSerialized, { error: undefined, isInline: false })
+        snapshotState.added++
+      }
+
+      return {
+        message: () => 'message a',
+        actual: '',
+        count,
+        expected: '',
+        key,
+        pass: true,
+      }
+    } else {
+      if (!pass) {
+        snapshotState.unmatched++
+        return {
+          message: () => 'message b',
+          actual: receivedSerialized,
+          count,
+          expected: expected !== undefined ? expected : undefined,
+          key,
+          pass: false,
+        }
+      } else {
+        snapshotState.matched++
+        return {
+          message: () => 'message c',
+          actual: receivedSerialized,
+          count,
+          expected: '',
+          key,
+          pass: true,
+        }
+      }
+    }
+  },
+})
+
 describe('TimeSeries', () => {
   const tMin = new Date('1970-10-01T00:00:00.000Z')
   const tMax = new Date('1971-02-01T00:00:00.000Z')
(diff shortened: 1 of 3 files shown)


src/algorithms/utils/TimeSeries.test.ts
```diff
@@ -91,7 +191,8 @@ describe('TimeSeries', () => {
         return interpolator(date)
       })
 
-      expect(result).toBeArrayOfSize(intervalCount).toMatchSnapshot()
+      expect(result).toBeArrayOfSize(intervalCount)
+      expect(result).toBeCloseToArraySnapshot()
```
abrie: Use the extension to compare the result with the expected snapshot.

src/algorithms/utils/TimeSeries.test.ts
```diff
@@ -1,6 +1,106 @@
 import { uniformDatesBetween, makeTimeSeries, updateTimeSeries } from './TimeSeries'
 import { interpolateTimeSeries } from '../run'
 
+import * as fs from 'fs'
+const { utils } = require('jest-snapshot')
+
+expect.extend({
+  toBeCloseToArraySnapshot(received: number[]) {
+    const { testPath, currentTestName, snapshotState } = this
```
abrie: Jest matchers are invoked with a context, which contains a mutable 'snapshotState' object and the test name. We need those in order to continue.

src/algorithms/utils/TimeSeries.test.ts
```diff
@@ -1,6 +1,106 @@
...
+    snapshotState._counters.set(currentTestName, (snapshotState._counters.get(currentTestName) || 0) + 1)
+    const count = Number(snapshotState._counters.get(currentTestName))
+    const key = utils.testNameToKey(currentTestName, count)
+    const expected = snapshotState._snapshotData[key]
```
abrie: Use the test name and the snapshot object to get a reference to the expected string.

src/algorithms/utils/TimeSeries.test.ts
```diff
@@ -1,6 +1,106 @@
...
+
+    /* If this isn't done, Jest reports the test as 'obsolete' and prompts for deletion. */
+    snapshotState._uncheckedKeys.delete(key)
```
abrie: Note the comment.

src/algorithms/utils/TimeSeries.test.ts
```diff
@@ -1,6 +1,106 @@
...
+    const tryParse = (str: string) => {
+      try {
+        return JSON.parse(str)
+      } catch (e) {
+        return []
+      }
+    }
+
+    /* If the expected field is blank, then this is probably a new snapshot. */
+    const expectedDeserialized = expected ? tryParse(expected) : []
```
abrie: Try to parse the expected string into, we assume, an array of floating point numbers. This commit is a hack, so no sanity checks.

src/algorithms/utils/TimeSeries.test.ts
```diff
@@ -1,6 +1,106 @@
...
+
+    const pass =
+      expectedDeserialized.length > 0 /* must check this because every() returns true if empty array */
+        ? received.every((_, idx) => Math.abs(expectedDeserialized[idx] - received[idx]) < 10 ** -2 / 2)
+        : false
```
abrie: Loop through the expected results and compare each float to the received results. The comparison function is equivalent to the default toBeCloseTo(), which checks for a difference of < 0.005.

src/algorithms/utils/TimeSeries.test.ts
```diff
@@ -1,6 +1,106 @@
...
+    const hasSnapshot = expected !== undefined
+    const snapshotIsPersisted = fs.existsSync(snapshotState._snapshotPath)
+    const receivedSerialized = JSON.stringify(received, null, 2)
```
abrie: Serialize the float array as JSON with some gentle formatting.

src/algorithms/utils/TimeSeries.test.ts
```diff
@@ -1,6 +1,106 @@
...
+          count,
+          expected: '',
+          key,
+          pass: true,
+        }
+      }
+    }
+  },
+})
+
```
abrie: Much of this code was lifted directly from Jest's `toMatchSnapshot()` implementation. Note that no useful error information is printed. That will need to be fixed.

src/algorithms/utils/TimeSeries.test.ts
abrie: Use the extension to assert the test.

src/algorithms/utils/__snapshots__/TimeSeries.test.ts.snap
```diff
@@ -1,19 +1,17 @@
...
-  4.205626732055542,
-  5.645209732882576,
-  6.888916295468725,
-  7.8873202652312076,
-  8.676811131585438,
-  9.322659276563021,
-  9.89015684624425,
-  10,
-]
-`;
```
abrie: Here is a default serialization of Jest's snapshot...

src/algorithms/utils/__snapshots__/TimeSeries.test.ts.snap
```diff
@@ -1,19 +1,17 @@
...
   2.6354594390679185,
-  4.205626732055542,
-  5.645209732882576,
-  6.888916295468725,
-  7.8873202652312076,
-  8.676811131585438,
-  9.322659276563021,
-  9.89015684624425,
-  10,
-]
```
abrie: This is Jest's default serialization of an array of floats.

src/types/jest.d.ts
```diff
@@ -11,10 +11,16 @@ import 'jest-extended/types'
...
+declare global {
+  namespace jest {
+    interface Matchers<R> {
+      // After toBeTruthy() we are certain that the matched value is not null.
+      // Modify the return type accordingly.
+      toBeTruthy(): NonNullable<R>
+    }
+
+    interface ChainedMatchers<R> {
+      toBeCloseToArraySnapshot(values: number[]): R
```
abrie: Keep tsc happy.

src/types/jest.d.ts
```diff
@@ -11,10 +11,16 @@ import 'jest-extended/types'
...
+    }
```
abrie: Keep `tsc` happy.

src/algorithms/utils/TimeSeries.test.ts
abrie: Use the test name and the snapshot object to get a reference to the expected string.

src/algorithms/utils/TimeSeries.test.ts
```diff
@@ -1,6 +1,106 @@
...
 import { interpolateTimeSeries } from '../run'
 
+import * as fs from 'fs'
+const { utils } = require('jest-snapshot')
+
+expect.extend({
+  toBeCloseToArraySnapshot(received: number[]) {
+    const { testPath, currentTestName, snapshotState } = this
+    snapshotState._counters.set(currentTestName, (snapshotState._counters.get(currentTestName) || 0) + 1)
+    const count = Number(snapshotState._counters.get(currentTestName))
```
abrie: This code was lifted from Jest's `toMatchSnapshot()` implementation. Not entirely sure what it's doing, but the test doesn't work without this.

src/algorithms/utils/TimeSeries.test.ts
ivan-aksamentov: Wow, facebook codez :)

src/algorithms/utils/TimeSeries.test.ts
ivan-aksamentov: @abrie Can this be parameterized?

This is a very silly way of doing comparison, because it assumes a particular scale of values.
For example if my expected floats are `0.001` (e.g. 0.1% mortality rate in a certain age group)  and received is `0.003` (0.3%) this will fail to detect it (false negative). Similarly, false positives for large values, but that is a concern for astrophysicists, not us.

I understand that this is how Jest does things. But what do I say our hospital administrators when their patient's mortality suddenly increases 3 times? That facebookz codez are cool and everybody just needs to chill? :)

abrie: Yes, this can be parameterized – like Jest's [`toBeCloseTo()`](https://jestjs.io/docs/en/expect#tobeclosetonumber-numdigits), which takes an optional second parameter 'numDigits`. Definitely something that should be incorporated here.

src/algorithms/utils/TimeSeries.test.ts
abrie: Never heard the term "facebook codez" before... but it does seem to be an fitting description.

PR #305, opened 2020-03-30
bash compatible .env.example file

kbobrowski:
## Description

String variables containing space should be surrounded by ",
e.g. WEB_APP_NAME_FRIENDLY="COVID-19 Scenarios".

Without " above command will cause following error in bash:
Scenarios: command not found

It is still correctly loaded from .env file by js, but if bash / zsh is set up to automatically parse env variables from .env file then above error will be displayed every time root directory is entered.

## Testing

Make sure that WEB_APP_NAME_FRIENDLY is correctly loaded from .env file when its value is surrounded by "


ivan-aksamentov:
Hi @kbobrowski, thanks. Why do you need this?

I mean our `.env` is only used by `dotenv` package in the build and by `docker-compose`.  What business does shell have here?

Is this just some of your plugins you want to unbreak?

I think quotation marks might either break dotenv or docker-compose, don't remember which one. But I clealy remember trying that with hostnames back in the day. So now I don't put quotes in there.

kbobrowski:
Yes it breaks my plugin (dotenv for zsh), but indeed dotenv js package will also look into .env file, not only env variables in the shell, and then it parses it correctly. I thought it would be harmless change, but if there is a chance it might break something / do not work across different platforms then it's better to leave it like it is right now

PR #308, opened 2020-03-30
validation of numberPoints in reducer

### LINKED ISSUE #307: editing number of points for mitigation curve may throw "Containment cannot be empty"
# 🐛 Bug Report

## How to reproduce

Steps to reproduce the issue:

 2. Remove Number of Points value with backspace
 3. Wait until mitigation curve disappears
 4. Type anything in number of points field (e.g. '5')
 5. Check if "Containment cannot be empty" is thrown

## 😯 Current Behavior

Error
Containment cannot be empty
Call Stack
 interpolateTimeSeries
  Home.js:216452:11
 updateTimeSeries
  Home.js:216605:98
 undefined
  Home.js:222994:62
 undefined
  Home.js:222907:125
 Immer.produce
  Home.js:46388:18
 undefined
  Home.js:222907:108
 undefined
  Home.js:146175:20
 undefined
  Home.js:146203:26
 updateReducer
  main.js:39378:22
 Object.useReducer
  main.js:40175:16

## 🤔 Expected Behavior

Mitigation curve should not be updated when number of points is outside desired range (5-100), which would ensure that it is never empty

## 💁 Possible Solution

containment.numberPoints can be null or outside the range 5-100, current Field component from formik does not enforce validity of it when editing the field with keyboard. This may affect other formik fields (haven't checked). Either some improvements in Field components are needed, or validity of the numberPoints value can be ensured in the reducer.

kbobrowski:
## Related issues and PRs

 - fixes #307 

## Description

formik Field updates field value even when outside specified
min-max range (e.g. when updating with keyboard, not with arrows).
This allows for null value of numberPoints, or for value outside
desired 5-100 range, which may lead to 'Containment cannot be
empty' error in algorithms/run.ts. Added aditional validation
of numberPoints value in the reducer.

Behavior of mitigation curve when editing number of points with keyboard (it will not change when number of points is not within 5-100 range).

## Testing

Try removing value of number of points with keyboard, or setting value which is not in the 5-100 range, mitigation curve should not change.

### DIFF
--- a/src/components/Main/state/reducer.ts
+++ b/src/components/Main/state/reducer.ts
@@ -90,11 +90,16 @@ export const scenarioReducer = reducerWithInitialState(defaultScenarioState)
 
   .withHandling(
     immerCase(setContainmentData, (draft, { data }) => {
+      let validNumberPoints = draft.data.containment.numberPoints
+      if (data.numberPoints && data.numberPoints >= 5 && data.numberPoints <= 100) {
+        validNumberPoints = data.numberPoints
+      }
+
       draft.scenarios = maybeAdd(draft.scenarios, CUSTOM_SCENARIO_NAME)
       draft.current = CUSTOM_SCENARIO_NAME
       draft.data.containment = {
-        reduction: updateTimeSeries(draft.data.simulation.simulationTimeRange, data.reduction, data.numberPoints),
-        numberPoints: data.numberPoints,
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
@kbobrowski Thanks!
We will be replacing this entire widget with another solution soon.

PR #313, opened 2020-03-30
Lowered euler step size

nnoll:
The integration of the ODE should be more accurate.

## Related issues and PRs

None

## Description

Changed the step size/day. Removed hard-coded instances of this assumption.

The ODE integration and the line plot which assumed a 4 times/day sampling.

## Testing

Performance could be impacted. Please test.

### DIFF
--- a/src/algorithms/model.ts
+++ b/src/algorithms/model.ts
@@ -77,13 +77,13 @@ export function getPopulationParams(
   }
 
   // Compute age-stratified parameters
-  const total = severity.map(d => d.ageGroup).reduce((a, b) => a + ageCounts[b], 0)
+  const total = severity.map((d) => d.ageGroup).reduce((a, b) => a + ageCounts[b], 0)
 
   let severeFrac = 0
   let criticalFracHospitalized = 0
   let fatalFracCritical = 0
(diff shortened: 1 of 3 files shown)


ivan-aksamentov:
@nnoll I did not notice anything. Do we do more looping now? Should I look at something specific?

ivan-aksamentov:
Do you want me to add a separate collapsible card for Advanced params?

nnoll:
I personally don't think we should expose this parameter. It is a technical detail at best and could go away in the future if we change how we integrate the ODEs. I mostly wanted you to test if the performance noticeably changed. I'm on a relatively beefy desktop so I'm not the ideal tester.

rneher:
I don't notice a change. We are now doing 5x more compute. this suggests that solving the ODE really isn't the issue. I suppose all the other linting changes are of the good kind?

nnoll:
@rneher. We had a bit of overhead that I removed that caused some noticeable. We would append all time-points to one large array, and then filter these out for plotting. It made more sense to just keep track of the current state of the ODE and then append the result every day. Less useless computation.

rneher:
Yes, filtering this down right away is good.

PR #316, opened 2020-03-30
compare just number of points in callback

### LINKED ISSUE #315: bug after changing number of points for mitigation - first move of point bounces back
# 🐛 Bug Report

## How to reproduce

Steps to reproduce the issue:

 2. Change number of points for mitigation curve
 3. Move one point
 4. Check that point bounced back to initial position --> error
 5. Move one point again
 6. Now it's fine

## 😯 Current Behavior

Point bounces back to initial position after changing number of points for mitigation curve

## 🤔 Expected Behavior

Point should stay in place after moving it

## 💁 Possible Solution

The error is caused by debounced callback in Main.tsx - newParams still hold old mitigation curve, simple workaround would be to use mitigation data from allParams which is up to date, and just update number of points from newParams

kbobrowski:
## Related issues and PRs

 - fixes #315 

## Description

fixes bug which results in point on mitigation curve returning
to initial position after number of points has been changed

behavior of mitigation widget after changing number of points

## Testing

check if moved point on mitigation curve stays in position after number of points has been changed

### DIFF
--- a/src/components/Main/Main.tsx
+++ b/src/components/Main/Main.tsx
@@ -152,9 +152,14 @@ function Main() {
     if (!_.isEqual(allParams.simulation, newParams.simulation)) {
       scenarioDispatch(setSimulationData({ data: newParams.simulation }))
     }
-    // NOTE: deep object comparison!
-    if (!_.isEqual(allParams.containment, newParams.containment)) {
-      scenarioDispatch(setContainmentData({ data: newParams.containment }))
+    // NOTE: compare only value controlled by formik!
+    if (allParams.containment.numberPoints !== newParams.containment.numberPoints) {
+      scenarioDispatch(setContainmentData({
+        data: {
+          reduction: allParams.containment.reduction,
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
Hi @kbobrowski thanks. We will not accept any more fixes to the mitigation plot as it is about to replaced with completely different widget. Right now these fixes only create rebasing noise.
Part of the motivation for replacement is that there are too many bugs :)

I will close this, but thanks for your efforts, much appreciated.

kbobrowski:
Hi @ivan-aksamentov thanks for the info about new widget being prepared, started working on fixing current version since I'm interested in forecasting required mitigation measures over long time frame, next enhancement I had in mind would be to enable modification of interval between points instead of absolute number of points, such that when time range is modified it would not change mitigation measures. But if it is supposed to be changed I will skip it for now :)

Do you have working branch for it, or some more detailed description of your idea?

PR #319, opened 2020-03-31
Fix-tests / Update snapshots for tests broken by changes in #313.

abrie:
## Related issues and PRs

Fixes broken tests related to #313

## Description

An algorithm change in #313 broke two tests in  `model.test.ts`.

1. "model getPopulationParams produces the expected output in the default scenario."
2. "model › evolve › produces the expected output in the default scenario after 5 evolutions."

Both tests rely on snapshots. This PR updates snapshots to the model's new behavior.

Unit test in `src/algorithms/model.test.ts`

## Testing

`yarn test model`


ivan-aksamentov:
@rneher  @nnoll The changes in what we get as results are unexpectedly quite dramatic here before and after #313. 

What you see in the diff are the differences between an old and a new snapshot. Snapshots were created from the results returned by one of the previous versions and saved to a file (`.snapshot`). During tests we compare returned results to the snapshot, they should match. After #313 we got a mismatch. @abrie ran `yarn test -u` or similar to update snapshots with new values and we are about to commit this as a new "ground truth". You can see the diff.  Red are the values that we expected before and green are what we actually get now.

Could you please double check for a potential unexpected breakage?

This test is what was failing:
https://github.com/neherlab/covid19_scenarios/blob/master/src/algorithms/model.test.ts#L95-L119

nnoll:
Yes, I expect very large differences in this test as it is written as that test advances 5 steps, no matter what the step size is. The previous step size was 1/4 of a day so that we advanced 1 1/4 days. The step size is now .05 which means we walk 1 step of the original algorithm. Comparing the results on the basis of days instead of steps solves this problem.

abrie:
@nnoll Since the step size is now 5 times smaller, I was curious what would happen if evolve was run 5 times longer (25 steps). [Here is a structure describing the differences](https://gist.github.com/abrie/d106c157d134968aae17a03361bcb44d#file-result-json) – the structure is same shape as the model's output, but the values are the absolute difference between the 0.25 and 0.05 versions of the model.

I wonder if this type of comparison is useful for model development?

rneher:
One way to approach this more systematically is to abstract the ODE integration into a function that integrates up to a certain time irrespective of the time step. then the test could run for a specified number of days and test for "closeness" rather than exact match. This should be robust to changes in integration time steps and integration algorithm.

rneher:
But looking at the numbers, the result is consistent with reducing the time by a factor of 5. (only the first exposed category is really seeing any influx when integrating for a few hours. A robust test would need to integrate for one week or so to have flux into all the categories.

PR #323, opened 2020-03-31
feat: add interval-based mitigation widget [WIP]

### LINKED ISSUE #14: Specify mitigation scenarios as intervals
In some situations, it might be more convenient to specify the mitigation scenarios as time intervals:
```
--2020-03-01                    100%
2020-03-01 -- 2020-03-15         50%
2020-03-16 -- 2020-03-31         75%
2020-04-01 -- 2020-05-31         60%
2020-06-01 --                    50%
```
The data structure for this would work just about the same way as a time series of adjacent intervals, the main difference would be UI. Need to think about how to enforce that intervals tile time.

ivan-aksamentov:
## Related issues and PRs

 - Aims to resolve #14 

## Description

When done, this will replace the existing mitigation widget and time-series with completely new ones.

Mitigation will be implemented with named intervals, each having an end and start date as well as a mitigation strength. Each interval represents an infection control intervention (distancing, ban on mass gatherings, school closures, lockdown, etc.)

Mitigation widget, application's main form state (context) and the algorithm

## Testing

Hopefully


rneher:
I added a conversion of the intervals to a time series. this is now interpolated linearly as before and it works in principle:
image
But it errors out when you delete an interval.

kbobrowski:
Added draft of mitigation chart, which enables interactive interval modification by dragging and resizing bars with a mouse, if you like the idea I could integrate it with current mitigation table :)

rneher:
a quick list of things that I think are necessary

 -[] fix error on deleting a measure
 -[] add measures to plot as horizonal bars
 -[] add measures to scenario
 -[] sensible defaults

rneher:
I just added interval mitigation measures to the script that produces the scenario json:
https://github.com/neherlab/covid19_scenarios_data/commit/4aa0941998c2b4f6ed4c0503d1f494fbf4200e3c

this should help getting these mitigation measures read in and the widget populated for different scenarios

rneher:
I added a dummy mitigation measure line to the plot. The currently hard coded `mitigationIntervals` need to be passed into the plot component and then positioning and styles need adjusting.

alebrand:
Following discussion with @rneher, I plan to:
- [ ] create a google form to collect the mitigation measures from various countries/states (circulate with e.g. Twitter)
- [ ] define preset mitigation values for each mitigation measure
- [ ] write script to assign the mitigation values to each measure in each country (output: TSV and/or JSON?)
Any suggestions at this stage are welcome, thank you. I understand the output will end up in assets/data/scenarios/scenarios.json, but should I store the TSV in data/ folder rather?

alebrand:
As part of the "mitigation" measures, I think we should also consider the fact that countries increase the number of ICU beds or hospital capacity, and hence, those values should be timeseries rather than a fixed value. I could do this in a 2nd step in a similar way to the mitigation values (i.e. google form to collect this information, then update data/populationData.tsv

alebrand:
>     * [ ]  create a google form to collect the mitigation measures from various countries/states 
Something like this?
https://forms.gle/Y2KXh47iwVKPY7t4A

rneher:
@ivan-aksamentov thanks a lot, this work really well. We might want to add the name of the measure somewhere, but the basic functionality is great. I am not sure how this look when many such measures overlap, but we can play with the styles a bit.

rneher:
@alebrand Your form is great. Maybe slightly too many categories. This will get a little messy. If we would lump nursery/schools/universities into one. and then have shops/restaurants and soft and hard lockdown this is probably enough. But overall this is great.

alebrand:
> @alebrand Your form is great. Maybe slightly too many categories. This will get a little messy. If we would lump nursery/schools/universities into one. and then have shops/restaurants and soft and hard lockdown this is probably enough. But overall this is great.

Thanks for the feedback @rneher , I updated the form accordingly, feel free to share the link around!
https://forms.gle/Y2KXh47iwVKPY7t4A

FYI the results are available on this google spreadsheet, I will work on how to integrate it within the code asap.
https://docs.google.com/spreadsheets/d/1CBYOSblzNSydSgWLcgPEb3l1IXHF4CPphmYtvyPvGKA/edit?usp=sharing

ivan-aksamentov:
@rneher @alebrand 
Idea: in ideal world we would gather this data right in the app and send it to store somewhere, then analyze and incorporate into the scenarios. Just another idea for a backend-based solution (either server or serverless), we currently don't have resources for :)

Issue #312, opened 2020-03-30
ICUs = 0 does not plot data


ressel:
Cool tool. Thank you! 👍 

# 🐛 Bug Report
If number of ICUs is set to zero, no data is plotted.

![ICU_Zero_Issue](https://user-images.githubusercontent.com/11699383/77946308-60b64000-72c2-11ea-9fbb-567eebd1a44b.JPG)

## How to reproduce

Steps to reproduce the issue:

 1. Open the application in a browser
 2. Set ICUs to 0.
 3. Run

## 😯 Current Behavior
Simulated data is calculated.
Data (both simulated and real) is **not** plotted.
Popup shows values for both simulated and real data.

If you switch to linear scale though, the missing data is plotted.

## 🤔 Expected Behavior
Data is plotted for log scale as well.

## 💁 Possible Solution
Potential root cause:
With log scale, the y value for 0 is undefined.  So grey horizontal line for ICU limit 0 cannot be drawn.
Solution: Do not draw line for ICU limit in this case.  It is not necessary anyway.

## 🔦 Context
I am playing around with some more or less artifical data, like being on an isolated island with no ICUs at all.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Browser          | Firefox
| Operating System | Windows 10
|                  |

## Related
 - A change of Hospital Beds does not seem to have any effect on the simulation.  Is that expected?

ivan-aksamentov:
Confirmed. Thanks for the report.
Just to clarify, are you trying to simulate a scenario where there are no ICU beds in the region?

nnoll:
This is a log(0) bug. We should add a check if any bed number is set to 0 and just not show it if it is.

ressel:
Yes, exactly.Thanks for your quick reply.Keep up the good work.
-------- Original message --------From: Ivan Aksamentov <notifications@github.com> Date: 3/30/20  20:56  (GMT+01:00) To: neherlab/covid19_scenarios <covid19_scenarios@noreply.github.com> Cc: ressel <ressel@gmail.com>, Author <author@noreply.github.com> Subject: Re: [neherlab/covid19_scenarios] ICUs = 0 does not plot data (#312) 
Confirmed. Thanks for the report.
Just to clarify, are you trying to simulate a scenario where there are no ICU beds in the region?

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.
[
{
[...]

Issue #154, opened 2020-03-23
'Confirmed cases past 3 days'


noleti:
# 🙋 Feature Request

'Confirmed cases past 3 days' as currently plotted in the 'Cases through time' is likely intended to represent the measured number of infectious people (as color is the same as simulated Infectious). I assume the 3 days period is thus chosen based on the 'infectious period' = 3 standard setting. Would it be possible to adapt the aggregation period for measured infected if the user changes the 'infectious period'? That would then more appropriately match to the simulated curve of infectious (which is influenced by the period).

## 🔦 Context

I am trying to accomplish meaningful matching of observed vs simulated infectious for non-standard 'infectious period' values.

## 😯 Describe the feature

Adapt the aggregation window by infectious period

## 💻 Examples

Example: [1,2,3,4,5] cases were observed on day 1 to 5. Currently, they are aggregated to [1,3,6,9,10] regardless to the size of infectious period. For infectious period of 5, the following would be my proposed aggregated values [1,3,6,10,15]

## 💁 Possible Solution

Not sure if the 3-day aggregation is done on user side, or server. If done on the user side, the 'infectious period' should be used as window size. If the aggregation is done on the server side, this feature will be harder to implement.

rneher:
thanks, @noleti. Good point and you are right that this is meant to match the `infectious`. The time window is currently hard coded in the client-side javascript:

https://github.com/neherlab/covid19_scenarios/blob/master/src/components/Main/Results/DeterministicLinePlot.tsx#L111

But we would need to decide how to deal with non-integers and put otherwise sensible bounds.

noleti:
non-integers should be possible to approximate by using the decimal places as weight for the first value in the window. But that would change your logic somewhat, and likely not fit into your nice one-liner. The GUI currently does not allow to set non-integer values btw. The `i>2` check also seems to depend on `infectious period`

noleti:
Another related point I realized re: matching observed infected to predicted infectious: real-world infectious measurements are likely by symptom-driven testing factor 5-10 times lower than the yellow line. This might confuse people who try to match predicted and observed infectious.

As detailed in the severity assumptions, for some age groups only 5-15% of cases show symptoms (but are still infectious, I assume). I also assume that those without symptoms appear in the solid yellow Infectious line (while they are infectious, i.e. 3 days). 
[...]

PR #325, opened 2020-03-31
filter empty data points to improve auto axis scaling in plot

rneher:
## Related issues and PRs
This is related to #312  (already fixed) but catches more such problems. 
Also related to #154 

## Description
This PR abstracts our calculation of new cases and removes all empty data points from the graph. 

plot. 

## Testing
inspect output. biggest change is expected for Germany where the case counts include data starting Jan 1st.


rneher:
If we hook up this:
https://github.com/neherlab/covid19_scenarios/blob/streamline/filter-empty-data/src/components/Main/Results/DeterministicLinePlot.tsx#L105

to the infectious period, that would solve issue #154

PR #327, opened 2020-03-31
#281/Custom Jest snapshot matchers

abrie:
## Related issues and PRs

Applies to #281 

## Description

~~The first pass, #295, introduced a functional proof-of-concept. That code ~is~ was merged into master, but the implementation was hacky.~~

This PR introduces ULP based comparison functions and a custom snapshot matcher for arrays of numbers. The code was developed on this branch, then moved to the npmjs repository for re-usability, see [https://www.npmjs.com/package/@eirba/jest-ieee754](https://www.npmjs.com/package/@eirba/jest-ieee754).
 
Unit tests.

## Testing

Run unit tests.


ivan-aksamentov:
@abrie Why close? I see you added some changes? :(

ivan-aksamentov:
@abrie Are there any particular issues with this PR to be resolved? Something you want to discuss?

Let me know if it's ready for review.

abrie:
I think we should discuss whether this is still relevant. The single TimeSeries test requiring this functionality has been removed. All the other tests are working fine. Presently I see no advantage to incorporating this into test suite.

Quite a bit of study went into building this extension, but the branch captures the knowledge and can be revisited as required.

ivan-aksamentov:
@abrie Would be a shame to abandon this. This might become a nice library to be used in many scientific projects. How about making an npm package or even contributing it into the jest?

Many our tests deal with arrays or objects of floats, so this is probably useful still. Plus I am going to start increasing coverage. You will see that there are new checks in PRs to encourage that.

One additional feature I would like to see is the ability to define the sensitivity in units in the last place (ULP). This is the ultimate correct way of float comparison. I did this before in C++, and I am not sure if it is doable in JS, but I think it should be. Basically, the idea is to define the precision in units of spacing between adjacent floating-point numbers (ULP). But because unit size is not uniform (depends on value, the larger the vaue, the larger the spacing), it requires  some disassembly of float representation (IEEE754) with bitwise operations. If we'd have it, it would be a fantastic contribution to scientific JS community.

ivan-aksamentov:
@abrie Holy bonkers, is this ULP comparison for real? O_O
If it actually works, you should definitely package on npm and/or submit PR to jest. This is gold.

abrie:
@ivan-aksamentov I'm not entirely sure this is the correct way to do ULP comparisons, but the unit tests are slowly increasing confidence.

As I understand it, the basic technique of ULP is to reinterpret a double precision (64 bit) value as a 64 bit integer. This code works by using Typed Arrays, DataView and BigInt. Their application is very straightforward. Still, floating point math is mind bending, so moving cautiously. [Bruce Dawson's blog](https://randomascii.wordpress.com/category/floating-point/) has been very useful.

ivan-aksamentov:
@abrie This is great! Yes, that's approximately how it's done in C and C++, for example in Google Test:

https://github.com/google/googletest/blob/e3f0319d89f4cbf32993de595d984183b1a9fc57/googletest/include/gtest/internal/gtest-internal.h#L340-L353

Maybe you find some inspiration there as well.

However, I am not sure what are the assumptions for JS regarding floating point implementation. Are all modern browsers and Node.js guaranteed to have numbers in IEEE-754 format? It would be crazy if there are any other implementations, but we all know how JS could be...

abrie:
> However, I am not sure what are the assumptions for JS regarding floating point implementation. Are all modern browsers and Node.js guaranteed to have numbers in IEEE-754 format? It would be crazy if there are any other implementations, but we all know how JS could be...

@ivan-aksamentov I think the assumption is safe. The ECMA standard requires that the `Number` type conform to IEEE-745, specifically as a 64-bit double precision value: https://www.ecma-international.org/ecma-262/6.0/#sec-terms-and-definitions-number-value.

abrie:
@ivan-aksamentov Yes, I think this is pretty close to being ready. I'll package it for distribution as descibed. Thanks for the pointer to [tsdx](https://github.com/jaredpalmer/tsdx).

To answer your question about usage in a browser: I don't think any polyfill will be required. Typed arrays and ArrayBuffer are well supported including IE10+; with the exception of the `ArrayBuffer.isView` method (which is not used by this package).

Issue #326, opened 2020-03-31
Using JSURL in links


dkozar:
# 🙋 Feature Request
Due to URL encoding, currently generated URLs are long and messy.

## 🔦 Context

We could use JSURL (https://github.com/Sage/jsurl) which would make the links shorter and even nicer.

With being shorter I mean: more data could be squeezed into this type of URL, without worrying of links being clipped by some systems.

## 😯 Describe the feature

Whenever the user changes parameters, we would update the route (browser link) and the parameters would be "JSURL encoded".

## 💻 Examples

JSON:

`{"name":"John Doe","age":42,"children":["Mary","Bill"]}`

JSON + URL encoding:

`%7B%22name%22%3A%22John%20Doe%22%2C%22age%22%3A42%2C%22children%22%3A%5B%22Mary%22%2C%22Bill%22%5D%7D`

JSURL:

`~(name~'John*20Doe~age~42~children~(~'Mary~'Bill))`

## 💁 Possible Solution

I used the JSURL before. The setup is trivial, here's the code snippet: https://github.com/dkozar/treelogy-demo/blob/09360f38c8974a57198eac7b65b0cdbaf5209c62/src/demo/utils/UrlUpdater.js

Here's the live example (some demo): https://dkozar.github.io/treelogy-demo/?data=~(versions~(~(id~%27Version_1~nodes~(~(id~%27foo~data~%27return*20*27foo*27*3b*20)~(id~%27bar~data~%27return*20*27bar*27*3b)))~(id~%27Version_2~parent~%27Version_1~nodes~(~(id~%27bar~data~%27return*20*27bar_2*27*3b)~(id~%27baz~data~%27return*20*27baz*27*3b)))))

Note that above link contains data for the demo.

![image](https://user-images.githubusercontent.com/1078403/78048873-4471dc00-7372-11ea-8464-e5d77b34240c.png)

Also note that, if backward compatibility with the old link is needed (for legacy purposes), our logic could check if the received link is of the old or the new format, and the app would be able to turn the old format into the new one.

**I could implement this feature.**

## Related

 - https://github.com/neherlab/covid19_scenarios/issues/31
 - https://github.com/neherlab/covid19_scenarios/issues/11
 - https://github.com/neherlab/covid19_scenarios/issues/284#issuecomment-606706469

dkozar:
I've been playing with the implementation.

PR (work in progress): https://github.com/neherlab/covid19_scenarios/pull/329
[...]

PR #329, opened 2020-03-31
feat: Using JSURL for data serialization in links

dkozar:
## Related issues and PRs

 - https://github.com/neherlab/covid19_scenarios/issues/326

## Description
Proposing alternative URL serialization (more read here: https://github.com/neherlab/covid19_scenarios/issues/326).

JSURL format: http://localhost:3000/?~(population~(ICUBeds~80~cases~%27CHE-Basel-Stadt~country~%27Switzerland~hospitalBeds~698~importsPerDay~0.1~populationServed~195000~suspectedCasesToday~10)~epidemiological~(infectiousPeriod~3~latencyTime~5~lengthHospitalStay~4~lengthICUStay~16~overflowSeverity~2~peakMonth~0~r0~2~seasonalForcing~0.2)~simulation~(simulationTimeRange~(tMax~1598918400000~tMin~1580428800000)~numberStochasticRuns~0)~containment~(~(y~1~t~1580428800000)~(y~1~t~1582483200000)~(y~0.6068708388814913~t~1584537600000)~(y~1~t~1586592000000)~(y~1~t~1588646400000)~(y~0.8250332889480693~t~1590700800000)~(y~1~t~1592755200000)~(y~0.5966444740346205~t~1594809600000)~(y~1~t~1596864000000)~(y~0.9170705725699069~t~1598918400000))~current~%27Custom)

Comparison of the old (URL encoded) and new (JSURL) format:

image

Note: There was an issue with Dates, because current application is passing dates to serializer. I'm suggesting to convert all the dates to Epoch before passing it to serializer.

The app will continue to read the legacy URLs, however the newly created ones will be in the JSURL format.

## Testing

This is a POC. I'm planning to write a thorough unit tests. There is a problem with current unit tests - it seems that a lot of tests are failing (in master branch).

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "i18next-browser-languagedetector": "4.0.2",
     "immer": "5.3.6",
     "immutable": "3.8.2",
+    "jsurl": "0.1.5",
     "jszip": "3.2.2",
     "katex": "0.11.1",
     "lodash": "4.17.15",
--- a/src/algorithms/types/Param.types.ts
+++ b/src/algorithms/types/Param.types.ts
@@ -61,21 +61,6 @@ export interface AllParams {
 
 export type AllParamsFlat = PopulationData & EpidemiologicalData & SimulationData & ContainmentData
 
-// part of the application state to be persisted in the URL
-export interface PersistedState {
-  current: string
-  data: ScenarioData
-}
-
-// format of the application state persisted in the URL
-export interface PersistedStateDto {
-  current: string
-  population: PopulationData
-  epidemiological: EpidemiologicalData
-  simulation: SimulationData
-  containment: TimeSeries
-}
-
 interface EmpiricalDatum {
   time: Date
   cases: number
--- a/src/components/Main/state/serialization/DtoConverter.test.ts
+++ b/src/components/Main/state/serialization/DtoConverter.test.ts
@@ -1,5 +1,6 @@
-import { PersistedState, PersistedStateDto } from '../../../../algorithms/types/Param.types'
 import { fromDto, toDto } from './DtoConverter'
+import { PersistedState } from './types/PersistedState.types'
+import { PersistedStateDto } from './types/PersistedStateDto.types'
 
 const STATE: PersistedState = {
   current: 'CHE-Basel-Stadt',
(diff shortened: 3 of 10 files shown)


dkozar:
I fixed the browser freezing problem and wrote unit tests around the de/serialization:

image

dkozar:
This is now ready for a merge. Are we going for the better URL encoding? 😄

src/components/Main/state/serialization/encodeURIComponent.ts
```diff
@@ -0,0 +1,44 @@
...
+      containment: containmentConverted,
+      simulation: {
+        ...obj.simulation,
+        simulationTimeRange: simulationTimeRangeConverted,
+      },
+    }
+  } catch (error) {
+    throw new Error('encodeURIComponent: Error while parsing URL')
+  }
+}
```
jsmith, line 39: Do you think it would be possible to merge these serialize/deserialize functions and the serialize/deserialize functions in `jsUrl.ts` into a single deserialize function which accepts a `parse` function and a single serialize function which accepts a `stringify` function? I see that you have extra logic in the `jsUrl` serialize function to convert `Date` objects to numbers but I think that is the only difference? This probably isn't that big of deal as once I merged in my `io-ts` changes it will make serialization/deserialization much simpler.

dkozar: Yeah, we can massage it further afterwards. Currently there are 2 pairs of serializers because of the backward compatibility. Once we use only JSURL, this will be even simpler.

Dates are problematic, that’s why I would suggest to use timestamps starting from the DTO level.

nnoll:
I fear we have a bit of duplicated work going on with respect to this feature. Specifically, how does this square with PR #338? If these are both solving the same problem, we should begin coordinating resources.

dkozar:
@nnoll they are not solving the same issue. Mine is solving the serialization into a string that is appended to the URL. Theirs is solving the problem one level above, and that is making the types used in the object to serialize more robust.

There are conflicts, but that doesn’t mean they solve the same thing.

Second one will obviously need to rebase.

ivan-aksamentov:
@dkozar Hi Danko! This is an interesting feature. Not sure if we really need this, but it is visually appealing, definitely. For example when sharing via email.

Our users are non-technical mostly. Maybe they'd prefer a shorter links (compression)? I don't know.

For sure, links should be shareable on social media.

I will do something else with URLs right now https://github.com/neherlab/covid19_scenarios/issues/283, because this is top priority before a release in a few days. But we will definitely come back to the link format.

dkozar:
> @dkozar Hi Danko! This is an interesting feature. Not sure if we really need this, but it is visually appealing, definitely. For example when sharing via email.
> 
> Our users are non-technical mostly. Maybe they'd prefer a shorter links (compression)? I don't know.

Hey,

JSURL gives you the balance between compression and insight what's in the data (it's human-readable).
> For sure, links should be shareable on social media (is there a length limitation there?)

Probably there is some limit.
> I will do something else with URLs right now #283, because this is top priority before a release in a few days. But we will definitely come back to the link format.

Sure. Hopefully this will sort of some parts of our URL serialization, which seems very messy right now and hard to follow.

I created a PR to address this, please tell me what you think: https://github.com/neherlab/covid19_scenarios/pull/387/files

Maybe we could use parts of this PR.

Note: This PR just improves the data flow and separates concerns (to the best of my knowledge/professional experience). It also introduces some unit tests.

dkozar:
@ivan-aksamentov I just pulled the latest from upstream/master and it seems the de/serialization is commented out, why is that? Currently the application doesn't persist anything in the URL.

ivan-aksamentov:
@dkozar sorry, master is a bit broken now, someone have merged something wrongly.

ivan-aksamentov:
@dkozar  Can you please send your email on Spectrum, we will invite you to our Slack?
https://spectrum.chat/users/ivan-aksamentov
It's faster to talk there

dkozar:
OK here we go.

Please note that there are some issues related to strings sometimes being received from the backend (?) instead of dates.

image

This has a tendency to break the app and should be fixed **at the source** ASAP:

image

In order for it to work, I've wrapped some calls to Date.getTime() with try..catch. Once we mitigate the root cause, those helper functions could be removed.

As far as I manually tested, persistance works as expected. There are issues with some pieces of a form (like Mitigation section), but thee doesn't seem to be related to serialization. Also, there is the 1 second debouncing issue that currently exists because of the form problem.

Although the PR seems extensive, the crux is here:

Snag_50980424

So from now on, JSURL serialiser is being used.

Note that this PR doesn't cover introduction of URL version.

Currently the URL is like:

`/?datablob`

In later PR I'll address this so we'll have:

`/?d=datablob&v=1`

Unit tests are working corrrectly:

image

PS. Note that, in the future, if you want to change the data that will be persisted in the URL, you have to change only the `DtoConverter.ts` file. This is the single file dealing with rearranging the object and replacing dates with Epoch.

PR #331, opened 2020-04-01
Switch numeral to numbro

### LINKED ISSUE #158: Human-friendly number formats (bonus: i18n)
# 🙋 Feature Request

## 🔦 Context

Currently the numbers in plot tooltips, and possibly in other places are displayed as raw values.
for example:
 ```
   deaths: 2654451
```
which is completely unreadable.

## 😯 Describe the feature

We need to display those in a human-friendly manner

```
deaths: 2.654 M

```

## 💻 Examples

See above

## 💁 Possible Solution

 - [ ] Wee need to preserve a number of ***significant*** digits (non-zeroes). I think d3 formatting library can do this pretty well. 

 - [ ] Would be nice to also internationalize these strings. 

 - [ ] !!!! Under no circumstances the actual runtime numbers should be rounded, converted or affected in any other way, only modify the displayed strings.

 - [ ] Optionally, please also add a global toggle, that switches between raw and human-friendly display (no need to add a widget, just make it so that a boolean can be passed in)

Related:
https://github.com/neherlab/covid19_scenarios/issues/56

rcbevans:
## Related issues and PRs
Fixes #158 

## Description
Replace numeral, with numbro, which has more i18n support

Screenshot of Polish abbreviated values

Rendering of numbers on results charts, tables.

## Testing
Switch language to Polish, observe that numbers are abbreviated with locale appropriate abbreviations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "lodash": "4.17.15",
     "mathjs": "6.6.1",
     "moment": "2.24.0",
-    "numeral": "2.0.6",
+    "numbro": "2.2.0",
     "p-min-delay": "3.1.0",
     "papaparse": "5.1.1",
     "prop-types": "15.7.2",
--- a/src/components/LanguageSwitcher/index.tsx
+++ b/src/components/LanguageSwitcher/index.tsx
@@ -2,20 +2,20 @@ import React, { useState } from 'react'
 import { Dropdown, DropdownToggle, DropdownMenu, DropdownItem } from 'reactstrap'
 import i18next from 'i18next'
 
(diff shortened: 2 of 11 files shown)


gj262:
hey @rcbevans, it looks like non-humanized is not working with this PR:

Screen Shot 2020-04-02 at 8 09 35 AM

rneher:
yes, non-humanized is broken for me as well.

gj262:
I fixed the problem, I think, rebased, broke git, force pushed and I think the version up now is good to go. But given the problems we've had with this feature, can someone give it a quick review.
CC @rneher @ivan-aksamentov

PR #333, opened 2020-04-01
Feat/solve ode until

rneher:
## Related issues and PRs
#313 

## Description
this PR separates the logic of stepping the ODE forward from integrating to a specified time. This will facilitate transitioning to a higher order integration scheme and should make tests more robust.

the model, specifically `run.ts` and `model.ts`

## Testing
tests might break because new snapshots are needed.


rneher:
Ideally, we would separate evolve function into a derivative and increment step. This would increase our flexibility when updating the algorithm.

PR #337, opened 2020-04-01
Adding io-ts for Runtime Validation

jsmith:
## Related issues and PRs

 - Contributes to #31
- Contributes to #283

## Description
Adds runtime type checking to url deserialization and localStorage. IMO this is very useful as it allows developers to be confident that the data they are working with is in the expected format and provides better error handling. These changes could be used as the foundation for the unified type scheme describe in the issue above.

A few more things:
- Btw `fp-ts` is a peer dependency! They both have 0 dependencies too which is nice.
- I also fixed a type issue in `Main.tsx`
- In the future, we could remove the `LOCAL_STORAGE_KEYS` as strings can be typed checked. For example, change:
```typescript
LocalStorage.set(LOCAL_STORAGE_KEYS.SKIP_LANDING_PAGE, 'true')
```
to
```
LocalStorage.set('skip-landing-page', 'true')
```

Both are the exact same thing from the point of view of TypeScript.

Impacts the url serialization/deserialization and local storage.

## Testing
Added two test suites, one for local storage and another for url serialization/deserialization. Specifically, it tests that these correctly decode information (ie. when data is in the expected format) and correctly handle errors (ie. when data is not in the expected format).


src/components/Main/Disclaimer.tsx
```diff
@@ -63,7 +58,7 @@ export default function DisclaimerProps() {
               className="form-check-input"
               onChange={toggleChecked}
               checked={suppressShowAgain}
-              ariaChecked={suppressShowAgain}
```
jsmith, line 66: This was throwing a error/warning for me in the console!

src/components/Main/Main.tsx
```diff
@@ -60,14 +59,18 @@ async function runSimulation(
     return
   }
 
-  if (params.population.cases !== "none" && !isRegion(params.population.cases)) {
```
jsmith, line 63: I see this fixed the `none` issue however it introduced type errors. I've refactored the code to allow us to handle this situation and to have everything type check!

src/helpers/localStorage.ts
```diff
@@ -9,17 +13,58 @@ export const LOCAL_STORAGE_KEYS = {
...
+  if (value === null) {
+    return null
+  }
+
+  let obj: unknown
   try {
-    return value !== null ? JSON.parse(value) : value
+    obj = JSON.parse(value)
   } catch {
-    return (value as unknown) as T
```
jsmith, line 21: Previously we returned the value if parsing failed. I changed this to emit a warning and return `null` instead which I think is better behaviour!

jsmith:
Closing in favour of #338

PR #338, opened 2020-04-01
[WIP] Using io-ts for Data Runtime Validation

jsmith:
## Related issues and PRs

 - Contributes to #31 
- Contributes to #283 

## Description
Adds runtime type checking to url deserialization and localStorage. IMO this is very useful as it allows developers to be confident that the data they are working with is in the expected format and provides better error handling. These changes could be used as the foundation for the unified type scheme describe in the issue above. In doing so, I also refactored `Param.types.ts` to be entirely defined by `io-ts` which allows for easy compile time and runtime validation of parameters. 

A few more things:

- Btw `fp-ts` is a peer dependency! They both have 0 dependencies too which is nice.
- I also fixed a type issue in Main.tsx
- In the future, we could remove the LOCAL_STORAGE_KEYS as strings can be typed checked. For example, change:
```typescript
LocalStorage.set(LOCAL_STORAGE_KEYS.SKIP_LANDING_PAGE, 'true')
```

to
```typescript
LocalStorage.set('skip-landing-page', 'true')
```

Both are the exact same thing from the point of view of TypeScript.

Impacts the url serialization/deserialization and local storage.

## Testing
Added two test suites, one for local storage and another for url serialization/deserialization. Specifically, it tests that these correctly decode information (ie. when data is in the expected format) and correctly handle errors (ie. when data is not in the expected format).

Edit: Modified my description to include all of the information from #337 which I closed in favor of this PR.


dkozar:
A stumbled onto this too. Yes, it looks like the previous de/serialization didn't persist all the needed values in the URL, and was picking some from the initial state.

Also, it was not clear what we're sending in the URL. Your solution of separating `current` and `data` sounds good, I didn't like how it was packed in the same level before.

jsmith:
Thanks @dkozar! I agree, it wasn't very clear what was being serialized/deserialized which is one of the benefits of using `io-ts`. Sorry that this conflicts with your JSURL changes. If my changes are merged before yours I can help resolve the merge conflicts!

dkozar:
No worries, I'm not planing to merge JSURL changes anyway before others confirm that they like it.

However, my PR has other improvements in the area of separation of concerns:

-  the serializer shouldn't really fetch `window.location` and so on
- de/serializers should be fully testable as utility modules (data in - data out)
- de/serializers should really a single parameter (object which will be serialised), and not contain the logic of picking certain parts of the application state

https://github.com/neherlab/covid19_scenarios/pull/329/files

src/algorithms/types/Param.types.ts
```diff
@@ -1,54 +1,61 @@
...
-  populationServed: number
-  country: string
-  suspectedCasesToday: number
-  importsPerDay: number
-  hospitalBeds: number
-  ICUBeds: number
-  cases: string
-}
+import { TimeSeriesType } from './TimeSeries.types'
+import * as t from 'io-ts'
```
brunorzn, line 2: wrong import order

jsmith:
@dkozar those changes definitely make sense! I looked through your PR and see what you mean. Also as I noted in the issue I removed the `current` key/value pair from serialization. Upon deserialization I set the scenario name to 'Custom' which I think makes the most sense.

nnoll:
Just jumping in here to raise a question. We currently also use the param.types.ts to generate the initial conditions for all scenarios within the data directory. The prototype of our solution basically copied the params structure over to python via classes, but this was suboptimal for obvious reasons. 

Our current solution is to use a JSON schema to generate these classes on the fly to ensure the structures always remain sync'd. You can see a sketch within PR #355. We should coordinate on this as it seems like you are moving the types into io-ts, however they need to be shared with our python scripts. How do you think we proceed here?

ivan-aksamentov:
@jsmith  Jacob, we are preparing to go with JSON schema for serialization: https://github.com/neherlab/covid19_scenarios/issues/283

I love `io-ts`, but we have so much more stuff and across languages so we decided to go with a more complex and general solution.

That being said, if I am not able to make JSON schema quickly (we plan a release soon), then we will go with `io-ts` for some time before it's done. 

I keep this open.

jsmith:
Thanks @ivan-aksamentov, I completely understand. I've read through the PR linked above and see that the data generated in Python must then be imported within TypeScript which is difficult.

Do you plan to use `ajv` for validation of unknown data? For example, when parameters are deserialized from URL parameters, local storage or file imports?

This PR also introduced runtime validation of the current local storage implementation (ie. storing `lang`, `autorun-simulation`, `suppress-disclaimer`, etc). If you would like me to create a PR just for this feature let me know!

ivan-aksamentov:
@jsmith Can you please send your email on Spectrum? we will invite you to our Slack.
https://spectrum.chat/users/ivan-aksamentov

I want everything to be covered with JSON schema basically, as in https://github.com/neherlab/covid19_scenarios/issues/283.

Later, parts of it of it may go to OpenAPI schema, for example instead of local storage we will send though HTTP to backend.

ivan-aksamentov:
Now that we have JSON schema-based validation this is no longer needed. But thanks for the effort anyway!

Issue #284, opened 2020-03-29
"Expand" the results panel


ivan-aksamentov:
# 🙋 Feature Request

I want to discuss the possibility of "zooming in" to a particular parts of the application. In this case, the results card and results plot.

## 🔦 Context

"Results" panel is the most important one. It's a heart of the app. But depending on screen size there are usability issues.

Even on large screens:
 - the results plot occupies less than 1/4 of the screen real estate
 - all the legend fields are crammed
 - most important values are hidden in the tooltip

In terms of highlighting of the important data none of this makes sense

## 😯 Describe the feature

I imagine a small button that will bring up (animated)  the whole results panel and/or a separate section of it full-screen or almost full-screen. Sort of like the images of items on shopping websites. A large popup of a kind.

In this case, after user is happy with the parameters and no longer needs the "Scenarios" panel, they could expand the Results panel in and see the data outputs more clearly.

Additionally, it gives us a change to restructure the layout in the popup, becaue it can be different from the current one:
 - to show more data
 - to move graph values on a side, instead of in the tooltip
 - to allow to zoom in on a particular range of dates
 - etc.

## 💁 Possible Solution

We currently use a couple of popups already (disclaimer, export), so this could be implemented similarly. It just has to be bigger.

## Related

 - https://github.com/neherlab/covid19_scenarios/issues/234

ivan-aksamentov:
Alternative or a complementary implementation is to collapse the "Scenario" panel when no longer needed.

Even more aswesome is to move the layout from hardcoded bootstrap grid to something dynamic like 
https://github.com/STRML/react-grid-layout
with fdragable, resizeable and expandable/zoomable panels

rcbevans:
I think having an "expand" on each of the three result graphs is a very easy start to make exploring the results easier;  could start with that, then move to a more flexible/dynamic layout.

I'll work on an expand now.

rcbevans:
@ivan-aksamentov I have a prototype that would allow displaying the entire result card body, or individual charts 'maximized' in a modal dialog, launched by clicking the 'expand' arrows beside the title of the component to be expanded.

![Screen shot showing the expand icon by the results header](https://user-images.githubusercontent.com/2041260/77842919-aa7d1a00-714c-11ea-87b2-27d8b8e12761.png)

![Screenshot of results rendering in a modal dialog](https://user-images.githubusercontent.com/2041260/77842891-668a1500-714c-11ea-9e33-b3f6644b7a34.png)

[...]

PR #339, opened 2020-04-01
Use a help popover that better supports window.open

gj262:
## Related issues and PRs

Related to: #284 

## Description

This change switches from reactstrap tooltips to the unfortunately name 'awesome' popovers. It works around the problem of reactstrap tooltips not being able to locate the tooltip target in child windows. It does this by wrapping the target and content so it has no need to query the DOM to find the target.

As a POC I wrapped just the results controls in a new window and verified that the help works in both.

BUT:

- [x] the style is horribly broken
- [x] tests are probably broken

Help tips.

## Testing

Busted.

- [x] requires a battery of cross browser and OS tests before (any hope of) adoption


ivan-aksamentov:
@gj262 Also need to make sure that Safari still works

gj262:
> @gj262 Also need to make sure that Safari still works

Yeah @ivan-aksamentov, assuming I can get the styling fixed, I'll put my saucelabs account to good use doing a battery of OS/browser tests. Come to think if it, I will first do a bit of that to ensure the new look works across OS/browsers.

gj262:
Ok! This is working now if we want the change to allow react-new-window to work. I verified the help behaves as expected on:

Windows 10
	Edge 80 - PASS

MacOS Catalina
	Chrome 80 PASS
	Firefox 74 PASS
	Safari 13 PASS

@ivan-aksamentov @rcbevans

rneher:
works well for me.

PR #334, opened 2020-04-01
Even more eslint fixes for #101

patrikvarga:
## Description

More `eslint` fixes.

## Related issues

Contributes to https://github.com/neherlab/covid19_scenarios/issues/101

## Impacted Areas in the application

None.

## Testing

Build, test & run.

PR #352, opened 2020-04-01
#336/Fix broken unit tests caused by react-children-utilities dependency.

abrie:
## Related issues and PRs

#336  

## Description

Fixes broken unit tests caused by [`react-children-utilities`](https://github.com/fernandopasik/react-children-utilities) dependency.

`FormHelpButton` uses the label attribute to populate the help box. If the label is a React node then the text is difficult to access programmatically. #336 introduces a dependency, `react-children-utilities` for the 'onlyText' method to extract the text from React nodes. But the dependency breaks the unit tests because the ESM module format is not supported by Jest (see open issue https://github.com/facebook/jest/issues/4842).

This PR creates a `HelpProps` interface to communicate the help label and text to the `FormHelpButton`. This solution is slightly more verbose than extracting text from the 'label' attribute. But it is more explicit and breaks no existing behavior.

This changes a lot of files, but in a predictable and uniform way.

Both the UI and unit tests are affected by these changes.

## Testing

`yarn test`


package.json
```diff
@@ -82,7 +82,6 @@
     "prop-types": "15.7.2",
     "raf": "3.4.1",
     "react": "16.13.0",
-    "react-children-utilities": "2.0.12",
```
abrie, line 85: Remove dependency.

src/components/Form/FormHelpButton.tsx
```diff
@@ -12,13 +10,21 @@ function safeId(id: string) {
   return id.replace('.', '-')
 }
 
+export interface HelpProps {
+  label: string
+  text: string
+}
```
abrie, line 16: This interface is passed into the 'help' attribute for elements that need a help button. Having `label` in this interface removes the need to try and extract label text from the the 'label' attribute on the element.

src/components/Form/FormHelpButton.tsx
```diff
@@ -12,13 +10,21 @@ function safeId(id: string) {
...
+
+export function help(label:string = "", text:string = "") : HelpProps {
+  return { label, text }
+}
+
```
abrie: This is a helper function to create the HelpProps object.

src/components/Form/FormHelpButton.tsx
```diff
@@ -36,8 +42,8 @@ export default function FormHelpButton({ identifier, label, help }: FormHelpButt
       <UncontrolledPopover placement="right" target={safeId(identifier)} trigger="legacy" hideArrow>
         <Card className="card--help">
           <CardBody>
-            {label && <h4>{onlyText(label)}</h4>}
-            <p>{help}</p>
+            <h4>{help.label}</h4>
```
abrie, line 45: Now we no longer need to invoke `onlyText`. We just insert the contents of the 'label' field from the HelpProps object.

src/components/Main/Results/ResultsCard.tsx
```diff
@@ -106,7 +107,7 @@ function ResultsCardFunction({
             {t('Results')}
           </h2>
         }
-        help={t('This section contains simulation results')}
+        help={help(t('Results'), t('This section contains simulation results'))}
```
abrie, line 110: This pattern is repeated throughout all elements that have a 'help' attribute. The attribute receives a HelpParams object, created by the `help(title, text)` helper.

src/components/Main/Scenario/ScenarioCard.tsx
```diff
@@ -40,7 +41,7 @@ function ScenarioCard({ severity, scenarioState, errors, touched, setSeverity, s
     <CardWithDropdown
       identifier="scenarioName"
       label={<h2 className="p-0 m-0 d-inline text-truncate">{t('Scenario')}</h2>}
-      help={t('Combination of population, epidemiology, and mitigation scenarios')}
+      help={help(t('Scenario'), t('Combination of population, epidemiology, and mitigation scenarios'))}
```
abrie, line 44: Here is an example of a label that renders a React node. Note that we pass both a label and help text to the help() method, which generates a HelpParams object. The label is duplicated in both the label= attribute and the help() function.

src/components/Form/FormHelpButton.tsx
```diff
@@ -12,13 +10,21 @@ function safeId(id: string) {
   return id.replace('.', '-')
 }
 
+export interface HelpProps {
+  label: string
+  text: string
```
gj262, line 15: Hi @abrie could `text` be changed to `content` and be `string | React.ReactNode`? When I changed this last, @ivan-aksamentov mentioned that he wanted to allow the content of help possibly be more than plain text.

abrie: @gj262 Yeah, sounds reasonable. PR Updated.

src/components/Form/FormHelpButton.test.tsx
```diff
@@ -35,8 +37,21 @@ describe('FormHelpButton', () => {
...
+    const { getByLabelText, findByText, queryByText } = render(
+      <FormHelpButton identifier="abc" help={help('def', <span>Hello, ReactNode</span>)} />,
+    )
+
+    fireEvent.click(getByLabelText('help'))
+
+    await findByText('Hello, ReactNode')
+    expect(queryByText('Hello, ReactNode')).toBeTruthy()
+  })
+
```
abrie, line 50: @gj262 Here's a unit test for setting HelpParams content to a `ReactNode`

src/components/Form/FormHelpButton.tsx
```diff
@@ -12,16 +12,16 @@ function safeId(id: string) {
...
 export interface HelpProps {
   label: string
-  text: string
+  content: string | React.ReactNode
 }
 
-export function help(label = '', text = ''): HelpProps {
-  return { label, text }
+export function help(label: string, content: string | React.ReactNode): HelpProps {
+  return { label, content }
```
abrie: @gj262 `text` renamed to `content`, and it allows React.ReactNode.

gj262:
Hi @abrie LGTM and thanks for tidying this up and fixing the test. However it looks like you missed one, the help on the Severity Card is not rendering with the preview.

abrie:
@gj262 Good catch! Thanks. Updated PR.
  
> Hi @abrie LGTM and thanks for tidying this up and fixing the test. However it looks like you missed one, the help on the Severity Card is not rendering with the preview.

ivan-aksamentov:
@abrie I would prefer to just transpile the package. In fact I already did it on another branch. Will backport right  now.

abrie:
Ok

ivan-aksamentov:
@abrie @gj262 
The "correct" fix, THE ONE AND ONLY, :DDD
is here
https://github.com/neherlab/covid19_scenarios/commit/0100bfa63c48c456faca36524999b4d3cd97e8e3
this fixes the transpilation of the module and nothing else.

Now, let's talk about the refactoring I see in the diff. Adding a `help()` wrapper is not entirely what I thought it will be. The idea was that we can put an arbitrary component inside the help box, including links, list of reference materials etc. So that it is more educative material that help box. Naturally, each box might be different.

@rneher Richard has never actually used any of this functionality, but I hope we can do it later still. 

What do you think folks? 

Tremendous push would be to scaffold a bunch of components somewhere, preferably in `.mdx` format and move all the help text over there, so that Richard could then expand on it without being constrained by the inline componentry thing we have now.

@abrie I only glanced over your PR, so let me know if I am absolutely wrong. It can easily be the case. In which case I trust Gavin @gj262 to merge this or something similar.

Feel free to create a coherent issue(s) to make it more transparent for the community.

gj262:
@ivan-aksamentov sounds good, your fix is similar to the one I tried in the window.open() fix. #334.

And I _like_ @abrie's approach of making the help popover label explicit and separate from the form label as it should be IMHO (separate concerns, etc.). So if the work in this PR can be resurrected I'll happily review and merge :)

abrie:
This was a poorly titled PR.

As @ivan-aksamentov noted, it is primarily a refactoring.

It fixed the unit tests as a side effect.

If the plan is a more elaborate help system, then these changes are probably worth exploring. The existing implementation is magical and could use a more explicit approach, i.e. separate concerns, etc (thanks @gj262).

For future enhancements, this PR provides an interface, [`HelpProps`](https://github.com/abrie/covid19_scenarios/blob/d2bdeaaa348406e874988d92c8e0223599ee4aa4/src/components/Form/FormHelpButton.tsx#L13), which would be the starting point for extension with SVG, mdx, etc: https://github.com/neherlab/covid19_scenarios/pull/352#discussion_r401934767.

Issue #342, opened 2020-03-28
Improved dependency management


tryggvigy:
- As a contributor I would like to quickly install the required dependencies to start working after cloning.
- As a maintainer I would like have explicitly declared, version frozen dependencies that I can audit for security issues.

I don't use python often. When I cloned the project and tried to generate date for the first time, I saw I had missing dependencies. I set up https://github.com/pypa/pipenv, which seems to be the tool of choice in the python ecosystem nowadays, and installing the missing dependencies.

I believe managing dependencies with `pipenv` could lower the barrier to contribution.

nnoll:
I usually keep everything within one pipenv because I'm lazy, but I do think it's good to be explicit about dependencies, especially if one wants to pull from the head of a repo. I've made a pipenv file that should allow one to just type pipenv install inside the data directory to accomplish this.

tryggvigy:
Thanks for considering. I think this is an improvement for new folks coming in to contribute. Especially those that do not work in python on a regular basis like myself

PR #353, opened 2020-04-01
Update data

rneher:
## Description
update data and make all path relative to base of the data directory

update data. only the tsv


rneher:
the case count json could now be generated on build with 
`python data/generate_data.py --output-cases src/assets/data/case_counts.json` in split seconds. 
should we add this json to gitignore and generate it? But this would require a python env

nnoll:
A python env is probably needed. It would resolve Issue #342

nnoll:
I'm also for ignoring case-counts.json (and scenario.json while we are at it) to reduce diff noise.

rneher:
scenarios take a while to compute and don't need recalc on a daily basis. but I agree in principle.

nnoll:
Merge this when you are ready Richard. I can include a Python environment when I finish the json schema overhaul

rneher:
I updated the scenario json for now. We can decide two to generate them later.

PR #355, opened 2020-04-01
[WIP] Type schema for data directory

nnoll:
## Related issues and PRs

This is intended to be a large PR to address Issue #283 from the data import side.

## Description

Currently, we deal with passing data into our app via code duplication: our input types within Typescript are duplicated by Python classes that ultimately create a scenarios.json which needs to be typed exactly the same as AllParams. This is suboptimal and will ultimately lead to friction as we change things. This PR will refactor this so there's only one source of truth.

Intended:
data/scripts/scenarios.py
data/scripts/model.py
src/algorithms/model.py
src/algorithms/types/*

Unintended/Additional:
Will be filled in as work continues on this branch.

## Testing

I'll be attempting to update tests as I go as this PR is intended to instill confidence in our data generation process.


ivan-aksamentov:
This was fixed in https://github.com/neherlab/covid19_scenarios/pull/354

ivan-aksamentov:
@nnoll As we discussed on Slack, creating a schema is a step to the right direction, but this particular library is not necessarily what I imagine it would be.


\<rant desc="old man yells at cloud"\>
<sub>
One particular worry is that [schemapi was last released: Oct 18, 2017](https://pypi.org/project/schemapi/), and the particular Github thing referenced in Pipfile is not hugely popular (59 stars. Even we have 15 times of that). This means the maintenance there is pretty weak. Also, referencing a library on master is probably a bit extreme. Just not what you do with the libraries that ensure correctness.
<br/>
What strikes me in the dynamic object generation approach, is that it provides no benefit compared to normal runtime validation + generated static classes, but it hides types from you, your text editor and linter and has this additional WTF effect when you see the `make_` factory functions everywhere.
</sub>
\</rant\>


How about we leave "hipster" solutions alone and proceed with:
 - class generation with [quicktype](https://github.com/quicktype/quicktype)
 - validation with either quicktype-generated `to_dict`/`from_dict` function or with  and/or "cannonical" [jsonschema](https://github.com/Julian/jsonschema)


I added the Python and Typescript generation here:
https://github.com/neherlab/covid19_scenarios/pull/390
Could you jump-in onto this branch and see if you can adapt it for the data tasks? Instructions are in the PR. Add your schema additions there too.

As a side note, could you please also eventually add targets for every data-related task to the `Makefile`?

nnoll:
@ivan-aksamentov I'm happy to do so, however in the hopes that I don't waste effort here I'm going to let this branch go stale transiently while the main app sorts out the schema and where it lives. And then I will resurrect this with that end solidified. Until then I feel I'm just wasting effort.

ivan-aksamentov:
@nnoll #390 is merged. I stole some of the schemas from here. Feel free to also salvaged whatever else is useful, close and re-PR.

nnoll:
Sounds good!

PR #359, opened 2020-04-01
🇮🇸 Improve Iceland parser

tryggvigy:
Previously the parser was reading from a data source that was generated from a Google Sheet file. This is brittle as you can imagine and the Iceland.tsv file already had missing data.

I've now created a repository in-between that normalizes the data. I've added the `recovered` data which has been made available this week. The parser is now more resilient to failures and I believe that changes I will need to make in the future can for the most part be done in https://github.com/tryggvigy/covid19_scenarios_is where I will execute unit tests and sanitize the csv files on every update (run every 1h)


nnoll:
This looks great. Thanks!

PR #363, opened 2020-04-02
Test merge

rneher:
## Related issues and PRs
#333 
#353 

## Description
trying to get a few additions, clean ups and small PRs that depended on each other in place. 

Model, scenarios, and deleted obsolete files

## Testing
might fail as ODE integration was abstracted.


rneher:
the snapshot comparison is failing. But this is expected. The current way of solving to ODEs should make this more robust going forward.

Issue #332, opened 2020-04-01
Custom age distributions


rneher:
## 🙋 Feature Request
We have had multiple requests for adjusting age distributions when local community doesn't match the country.

## 🔦 Context
Since severity is heavily age dependent, the burden on the health care system depends strongly on the age distribution of the population. 

## 💁 Possible Solution
Add a column to the severity table that is prepopulated with values of the country from a scenario but can be modified. 

- [x] add a column to the severity table
- [x] once modified, the value of the age distribution drop down should go to custom
- [x] currently, `state` contains a key for the age distribution which is read out in the `runSimulation` function. This should probably be replaced by the distribution proper (an array of 9 numbers)  
- [x] the URL should contain the values of the age distributions, rather than the country key
- [x] update the description of the severity card to reflect these changes.

gj262:
Hi @rneher in the project page for this issue the column is described as having numbers 0-100. However the data passed to run() are population counts per age segment. I would guess that you don't want the app to do that level of rounding. So. Should this column be raw numbers?

The column could have a total below it to help the user balance the numbers e.g. 
Your Total = 376000123
Country Population = 37600000

If the user chooses a province/state/county the column values could be extrapolated from the province/state/county population versus the country population to give them a saner starting point.

gj262:
I've created a branch and added the column to the table there. 
feat/custom-age-distribution.

gj262:
FYI, I'm working some more on this today.

gj262:
I think rather than modify the state.population.country value, and have that show up in the form field, it might be better to indicate visually that the age distribution has changed from defaults. Like:

![Screen Shot 2020-04-02 at 9 49 08 AM](https://user-images.githubusercontent.com/6854428/78276452-e7883a00-74c7-11ea-831e-c0c55291382e.png)

The * could be red and hoverable with "The age distribution for this scenario has been modified from the country defaults". A bonus feature would include a side by side comparison of the values in the info.

nnoll:
I worry about the asterisk, just because in principle you could modify the numbers to be substantially different from the initial condition, in which case the pointer is relatively meaningless. Personally I would vote for a simple change to "Custom" as a catch all state once you start fiddling with the numbers.

nnoll:
I would also vote for going with the basic numbers for the population, otherwise we'll need to impose a constraint on the column sum which seems inconvienent for users. For those who want to enter as fractions, the option still exists since this will be scaled by populationServed anyways.

[...]

PR #369, opened 2020-04-02
Custom age distributions

gj262:
## Related issues and PRs

Contributes to #332 

Still to do:

- [ ] styling - the numbers column is not large enough
- [ ] update card help text to reflect the feature
- [x] waiting on feedback re: comments in #332 
- [ ] number validation (positive integers)
- [ ] age distribution value -> 'Custom'

## Description

Allow customization of age distribution.

Age distribution values.

## Testing

- age distribution values set on initial load
- change when scenario or country drop downs change
- do not change when other form fields change
- edits are passed to run()
- edits persist in the URL

### DIFF
--- a/src/components/Main/Main.tsx
+++ b/src/components/Main/Main.tsx
@@ -26,13 +26,16 @@ import { schema } from './validation/schema'
 import { setContainmentData, setPopulationData, setEpidemiologicalData, setSimulationData } from './state/actions'
 import { scenarioReducer } from './state/reducer'
 import { defaultScenarioState, State } from './state/state'
-import { serializeScenarioToURL, deserializeScenarioFromURL } from './state/URLSerializer'
+import { serializeScenarioToURL, deserializeScenarioFromURL, deserializeSeverityFromURL } from './state/URLSerializer'
 
 import { ResultsCard } from './Results/ResultsCard'
 import { ScenarioCard } from './Scenario/ScenarioCard'
 import { updateSeverityTable } from './Scenario/severityTableUpdate'
 
 import './Main.scss'
+
+const scenarioStateFromURLHandling = deserializeScenarioFromURL(defaultScenarioState)
+const severityFromURLHandling = deserializeSeverityFromURL(severityData as SeverityTableRow[])
 
 export function severityTableIsValid(severity: SeverityTableRow[]) {
   return !severity.some((row) => _.values(row?.errors).some((x) => x !== undefined))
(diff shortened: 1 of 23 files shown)


src/components/Main/Scenario/SeverityTable.tsx
```diff
@@ -141,6 +140,16 @@ export interface SeverityTableRow {
...
+export type SeverityTableNumberFields =
+  | 'id'
+  | 'population'
+  | 'confirmed'
+  | 'severe'
+  | 'critical'
+  | 'fatal'
+  | 'totalFatal'
+  | 'isolated'
+
```
gj262, line 152: There is probably a better Typescript ™️  way to do this but that is beyond me.

src/components/Main/state/URLSerializer.ts
```diff
@@ -63,3 +65,40 @@ export function deserializeScenarioFromURL(initState: State): State {
...
+
+      // prettier-ignore
+      severity.forEach((row: SeverityTableRow) => {
+        ['id', 'population', 'confirmed', 'severe', 'critical', 'fatal', 'totalFatal', 'isolated'].forEach(
+          (fieldName) => {
+            if (fieldName in row) {
+              // eslint-disable-next-line no-param-reassign
+              row[fieldName as SeverityTableNumberFields] = parseInt(
+                `${row[fieldName as SeverityTableNumberFields]}`,
+                10,
```
gj262, line 89: There is probably a better Typescript ™️  way to do this but that is beyond me.

src/components/Main/Main.tsx
```diff
@@ -88,19 +101,55 @@ const isRegion = (region: string): region is keyof typeof countryCaseCountData =
   return Object.prototype.hasOwnProperty.call(countryCaseCountData, region)
 }
 
+enum Provenance {
+  DEFAULT = 'DEFAULT',
```
gj262, line 105: Maybe should be called StateLifeCycle or something.

rneher:
thanks so much. very useful.

gj262:
Ok. So. Having the value of the age distribution field update to "Custom" is proving problematic. And I'm not available tomorrow April/3 to work on it further. So if someone else can pick this up that would be great. Otherwise I'll be back on it on April/4.

The issues are:

- the URL gets too big with the severity table packed in there
- trying to manage state among URL/reducer/useState and across two state blobs (scenarioState, severity) is hard

gj262:
I'm going to close this in favor of #376 which avoids the issues of mixed useState/Reducer problems in this branch by putting age distribution into the scenarioState reducer so the interdependency between country selection and customized distribution can be dealt with in one sane place.

PR #375, opened 2020-04-03
[WIP] Feat/parameter uncertainty

nnoll:
## Related issues and PRs

See Discussion #269

## Description

Run has been changed to iterate over this array and compute an average of both the linear and quadratic terms over all stochastic realizations to allow for easy return of the mean and variance of outcomes.

## Testing

Currently there is an odd interaction between animation and the area plots. They require a few refreshes to be viewed. This needs to be looked at. Also, the error bars most likely aren't correctly computed at this stage as the susceptible goes all the way to 1. So I need to debug here first.

The parameter type has been updated to allow R0 to be a tuple value, specified by a lower and upper value. Validation will have to be updated to test for bad cases (high < low)

The data directory will need to be updated to output a range of possible R0 values.

Tests will need to be written for the affected areas.
## Requested help

The new form found at src/components/Form/RangeSlider is far from complete. I view it as a prototype that works on my screen (there are many hard-coded values here). My burgeoning React skills are at their limit to make this a robust component that we could use elsewhere potentially. I would like a pro to look at this.


nnoll:
Most glaring issues have been fixed. The only one I'm currently working on is the help box is now rendered below the other forms... And DeepScan is mad but many of those issues come from the rebase

nnoll:
In a seperate PR we should consider integrating the slider into the mitigation card

rneher:
just played with this a little more. The general direction is great. We need to think about two issues:

 * do we attempt to rebase this? Or try to move changes over in a new branch?
 * the simulation takes about 500ms for each trajectory. We can probably speed this up by moving to a higher order integration scheme and larger steps and optimizing the code. 
 * uncertainties need to be added to types and schemas.

nnoll:
I think it will be easier to close this an reopen from a new branch with some cherry-picked files.

PR #376, opened 2020-04-03
Custom age distributions II

gj262:
## Related issues and PRs

- Contributes to #332 
- Supersedes #369
- Punts URL serialization to a new issue yet to be opened.

### Still to do:

- [x] styling - the numbers column is not large enough
- [x] update card help text to reflect the feature
- [x] waiting on feedback re: comments in #332 
- [x] number validation (positive integers)
- [x] age distribution value -> 'Custom'

## Description

Allow customization of age distribution.

Age distribution values.

## Testing

- age distribution values set on initial load
- change when scenario or country drop downs change
- do not change when other form fields change
- edits are passed to run()
- ~~edits persist in the URL~~
- Age distribution country changes to Custom
- Cannot be negative
- Cannot be cleared

### DIFF
--- a/src/components/Main/Main.tsx
+++ b/src/components/Main/Main.tsx
@@ -15,8 +15,6 @@ import { run, intervalsToTimeSeries } from '../../algorithms/run'
 
 import LocalStorage, { LOCAL_STORAGE_KEYS } from '../../helpers/localStorage'
 
-import { CountryAgeDistribution } from '../../assets/data/CountryAgeDistribution.types'
-import countryAgeDistributionData from '../../assets/data/country_age_distribution.json'
 import severityData from '../../assets/data/severityData.json'
 
 import countryCaseCountData from '../../assets/data/case_counts.json'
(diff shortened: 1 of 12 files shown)


gj262:
Finally! A clean PR! Please review so I can merge before any new changes break it!

The gist of this change is:

- ageDistribution is added to scenarioState
- changes to it are tracked through that reducer
- and changes with population.country are reconciled
  - (doing this all in the same reducer made the change much simpler then prior useEffect() madness)
- I've simplified the use of react-grid in the SeverityTable
 - I could not get the table headers to not look ugly / truncate without doing this
 - I may have dropped some UX features from the prior implementation
 - If so lets put that in a new issue

@ivan-aksamentov @nnoll @rneher

nnoll:
Should we allow for non integer values here? It doesn't matter numerically/algorithmically but conceptually it's a bit odd to allow for fractional humans.

nnoll:
Other than that this looks amazing and I would vote we should merge this!!

gj262:
> Should we allow for non integer values here? It doesn't matter numerically/algorithmically but conceptually it's a bit odd to allow for fractional humans.

Thanks for catching that. validateInteger() is not validating that its an integer!

PR #380, opened 2020-04-03
fix(data): use current icu cases for the netherlands

hongaar:
## Description
This PR corrects the number of ICU cases for The Netherlands. It's now using the current counts instead of cumulative counts.

`data`

## Testing
ICU counts were pulled from https://www.stichting-nice.nl/covid-19/public/intake-cumulative (`intakeCumulative` field).

Now they use https://www.stichting-nice.nl/covid-19/public/intake-count (`intakeCount` field).

Open both sources to review differences. Also present in graphs on https://www.stichting-nice.nl/


noleti:
was there a prior issue/discussion suggesting that the data should not be cumulative? While I can understand that hospital/ICU occupancy as displayed can go down as well (eventually), the current data/README.md specifies all data must be cumulative. I commented on this as well in https://github.com/neherlab/covid19_scenarios/issues/154

@rneher What is the best solution here? hospitalized/ICU data cumulative, and then the app needs to apply some window or so, or make it explicit that hospitalized/ICU data should be current?

rneher:
Thanks for raising this. My take it that `cases`, `deaths`, and `recovered` are cumulative while the remaining fields a current. This would match the model output. I'll update the README.

Issue #110, opened 2020-03-22
Export to PDF, print layout and printing


ivan-aksamentov:
Additionally to the export to JSON and CSV, and so long awaited import features, we may also add export of parameters and results to PDF or a similar format (or multiple).

The use case: this would serve as a report generator for "busy executives". For example, scientists or medics would first tweak the simulator and then generate a PDF that is ready to be presented to their superiors, local politicians, medical services etc.
Also the same document might as well be printed right in the app. 

I think bootstrap have a way to setup print layouts separately. Alternatively we may just generate a page in background on the fly and then write it using HTML5 file API.

guandiyiyi:
@ivan-aksamentov, did we try solution of https://www.npmjs.com/package/react-pdf ? If not, I can give it a try and circle back if this solution work.

ceporche:
Hi @ivan-aksamentov I'd like to commit to this issue.

ivan-aksamentov:
We need to accelerate this as it is a priority feature now

noleti:
maybe completely unrealistic, but: automated text generation with great layout is of course possible with LaTex, with the main caveat that the .tex compilation cannot be done in the client-side app. Compilation on the server would require new code and is not without security issues (we would need good validation etc). Maybe someone knows an API that will compile for us? Or (for academic end users) we provide a .zip with .tex+images (and/or suggest overleaf to compile)>

rneher:
I like the tex idea and overleaf might be a solution (we use it a lot). But we now have a pretty good print-solution which will be good enough for the next milestone.

PR #386, opened 2020-04-03
Adding first version of a pdf export

ceporche:
## Related issues and PRs
Following this [issue](https://github.com/neherlab/covid19_scenarios/issues/110): We want the user to be able to download a PDF containing the scenario parameter and its result.

## Description
This PR uses [jspdf ](https://github.com/MrRio/jsPDF)to generate a pdf file from the html content of the page.
For now, table and charts are treated as image using [dom-to-image](https://github.com/tsayen/dom-to-image). As such table text is not selectable, in a future version table should have a special treatment to be selectable.

Import of the two libraries jspdf and domtoimage
An extra button have been added in the ExportSimulationCard and an extra function in exportResults.ts.
Ids have been added in ResultsCard and ScenarioCard component to retrieve their content.

## Testing
Run a scenario and click on export button, once the dialog is displayed click on the download button for pdf file. Open the downloaded file.
**File Example:** [covid.params.results.pdf](https://github.com/neherlab/covid19_scenarios/files/4427528/covid.params.results.pdf)


ivan-aksamentov:
@ceporche Thanks for the contribution! This is one of our priority features, because our users often ask about it.

Note that we don't have to preserve the exact looks of the app, especially the parameters. However, plots is an important part to get right.

We will need some additions that are not in the app: for example, in PDF there will be no tooltips in the results plot, so no actual numbers are visible. These numbers are the main result, and without them the PDF does not make sense at all.

There should probably be some sort of a table, with data similar to what we export in .tsv file, but, of course with reduced granularity.

Also, where appropriate, does it make sense to use something more react-specific, like `react-pdf` or similar?

rneher:
Thank you, this is super very useful. What would be the best way to integrate a bit of a preamble (like rendering a bit of markdown that gives a guide to interpretation and caveats)?

tryggvigy:
I should have looked closer through the issues/PRs because I started working on this today as well. Maybe we can join our efforts to come up with the best solution. [Here is my commit](https://github.com/tryggvigy/covid19_scenarios/commit/7683cab5f94a1cb5bd8c685118594cd13f6dd0c3#diff-2799016b3fc04b0c5f4eed32e2a69f64R69-R72).

@ceporche does `dom-to-image` allow for any post-processing of the document?
I used `html2canvas` which can modify the document after cloning it: https://github.com/tryggvigy/covid19_scenarios/commit/7683cab5f94a1cb5bd8c685118594cd13f6dd0c3#diff-2799016b3fc04b0c5f4eed32e2a69f64R69-R72. I believe this is a good place to add any extra markup text that will only be present in the PDF document.

rneher:
thank you guys for working on this. Just a bit of context: The reason why I think some preample/additional text is important is the following. These pdf reports might serve as policy advice and will be distributed without the context of the about page and other background. I feel it is hence important to include such information. 

Beyond that, I think we should aim for a layout that the population and epidemiological parameters on top. The mitigation intervals are actually shown in the graph, so they could move to the bottom or be otherwise deemphasized.

happy to touch base to discuss this feature further.

tryggvigy:
@rneher @ceporche I think we should embed a link to the simulator with the parameters in the report.

Two things to consider:

- The JavaScript dependencies used to generate such a report are [quite heavy](https://bundlephobia.com/result?p=jspdf@1.5.3) and can slow the simulator down if loaded up front. We should consider lazy loading the code only once the PDF download button has been pressed.
- Generating the report itself can take several seconds. We should disable the button after first click to prevent the expensive duplicate work and provide a spinner to indicate to the users that something is happening. Maybe it is possible to move some of the work to a WebWorker. I haven't looked into that. Not sure if it's possible since generating the report relies heavily on access to the DOM

ceporche:
Awesome! Thanks for all the feedback those are great points!

@tryggvigy Sorry you had to do double work, I'd be more than happy to join forces 😊
Let me see if I can sum up what's left to be done, this way it'll be easier to split work and notify eachother on what we're working on (let me know if I forgot anything):

**For the tech investigation**
> does dom-to-image allow for any post-processing of the document?
html2canvas can modify the document after cloning it: tryggvigy@7683cab#diff-2799016b3fc04b0c5f4eed32e2a69f64R69-R72. I believe this is a good place to add any extra markup text that will only be present in the PDF document.

@tryggvigy No, I don't think it does and this could be super useful indeed.
- [ ] We should investigate https://react-pdf.org/advanced#on-the-fly-rendering to generate the PDF.

**For the feature**
- [ ] There should probably be some sort of a table, with data similar to what we export in .tsv file, but, of course with reduced granularity. 
@ivan-aksamentov Could you be more specific on the expected granularity?
- [ ] We should add a **preamble text** 
@rneher @nnoll Could you tell us which text exactly you have in mind?
- [ ] We should embed a **link to the simulator** with the parameters in the report.
- [ ] We should consider **lazy loading** the code only once the PDF download button has been pressed.
- [ ] We should **disable the button after first click** to prevent the expensive duplicate work and provide a spinner to indicate to the users that something is happening.

@tryggvigy I'll be starting with the reactjs investigation as this can have quite a lot of impact on the rest.

tryggvigy:
I think that sums it up.

> @tryggvigy I'll be starting with the reactjs investigation as this can have quite a lot of impact on the rest.

@ceporche go for it.

ceporche:
> @ceporche go for it.
Unfortunately it doesn't seem the way to go as [there is a dependency issue](https://github.com/diegomura/react-pdf/issues/670) with core-js 3 and previous version of babel-runtime used in reactpdf dependency. [A PR is opened on the react textkit repo](https://github.com/react-pdf/textkit/pulls) to remove the old babel-runtime dependency but it hasn't been merged for several months. I notified the repo author [here](https://github.com/diegomura/react-pdf/issues/670) with no answer so far.

Issue #346, opened 2020-04-01
True browser history


dkozar:
# 🙋 Feature Request

### History of the left panel (parameters)

Currently we're pushing to browser history only when user clicks the "Run" button:

```js
window.history.pushState('', '', `?${queryString}`)
```

I'd expect that the URL gets updated each and every time the user touches the "Parameters" section.

Additionally, when the user uses "Back" browser button, the previous values in the form should be filled in. This means that "sliders" in the Mitigation sections should move to their previous positions:

![image](https://user-images.githubusercontent.com/1078403/78168085-4bb6e980-7447-11ea-9017-5688e066e64a.png)

### History of the right panel (graphs)

I'd also expect that, if I use the browser Back button, the "previous graph" gets loaded.

Currently we're not executing the algorithm unless the user explicitly clicks the Run button. Is this because of performance concerns?

## 🔦 Context

It's something that users expect to see.

## 😯 Describe the feature

We should use React router and do all the changes in our app indirectly via the URL.

This means that, when the Run button is clicked, we would only update the URL (without calling the chart rendering). Then, the React Router would react on the URL change, deserialize the URL into the state and run the algorithm. 

## Related

 - https://github.com/neherlab/covid19_scenarios/issues/31

thoo1:
going to take a shot at fixing this

PR #387, opened 2020-04-03
fix: Improving the state serialization to URL

dkozar:
## Related issues and PRs

- https://github.com/neherlab/covid19_scenarios/issues/283
- https://github.com/neherlab/covid19_scenarios/pull/329

## Description

Current de/serialization is messy and I feel we're not in control. Serialization methods are communicating with browser history and window.location. This makes those methods hard to test.

Additionally, the `params` property is being passed into serialization method, whereas in reality it wasn't really needed. I've removed it from the serialization methods altogether.

In order to separate the concerns, I've split the serialization process into 3 layers:

1. communicating with browser history and window location
2. repacking the state to DTO that will be serialized
3. actual serialization of the DTO to string

I wrote unit tests for each layer. I believe this is puts us back in control.

Note: The functionality didn't change, so if you merge it there will be no visible change. But the serialization process will be easier to follow, and changes to each level will be trivial ([unified type schema](https://github.com/neherlab/covid19_scenarios/issues/283), [JSURL](https://github.com/neherlab/covid19_scenarios/pull/329))

## Testing

Load the app in a browser, change parameters, reload the page and check if all the parameters are properly persisted.


jsmith:
I see that you added back the `params` parameter. I'm pretty this data is needed since the data in `scenarioState` is actually stale for about 1 seconds after changing a parameter. You can test this by changing a parameter and quickly pressing `enter`!

src/algorithms/types/Param.types.ts
```diff
@@ -61,6 +61,21 @@ export interface AllParams {
...
+}
+
+// format of the application state persisted in the URL
+export interface PersistedStateDto {
+  current: string
+  population: PopulationData
+  epidemiological: EpidemiologicalData
+  simulation: SimulationData
+  containment: TimeSeries
+}
```
jsmith, line 77: Since you are adding this I think it would be a good point to address the fact that `current` is irrelevant. What do you think we should do?

dkozar:
> I see that you added back the `params` parameter. I'm pretty this data is needed since the data in `scenarioState` is actually stale for about 1 seconds after changing a parameter. You can test this by changing a parameter and quickly pressing `enter`!

Let's do something about it! 😃 

I don't think there should be any stale data. Are there any performance concerns with serialization? I'm not sure (please confirm/deny), but running the algorithm is a performance bottleneck, and not the serialization? Could we mentally separate these two?

The object we're serializing is not huge. I've never seen performance problems with this object size.

I think along these lines:

- we could/should serialize (and update the URL) upon each parameter change (whenever the user touches parameters)
- we should call the actual algorithm only after the debounced interval

src/components/Main/Main.tsx
```diff
@@ -71,9 +71,10 @@ async function runSimulation(
   const caseCounts: EmpiricalData = countryCaseCountData[params.population.cases] || []
   const containment: TimeSeries = intervalsToTimeSeries(params.containment.mitigationIntervals)
 
-  // serializeScenarioToURL(scenarioState, params)
```
jsmith, line 74: This is the line I mean! I don't think we can remove `params` since it contains the non-stale data being used.

dkozar: I understand, just commented above. To me it seems we're doing something wrong and should look at the problem from another angle.

dkozar:
Please note that this "constant serialization" to URL (upon each param change) is needed later. I'm about to address the "true history" issue: https://github.com/neherlab/covid19_scenarios/issues/346

So, all the changes need to immediately be pushed to browser history, and the user could use the browser Back button to get the previous set of values. But that doesn't mean that - if the user does this quickly - we'll call the algorithm each and every time and kill the browser. The call to the algorithm will stay debounced!

If we're not doing it this way (i.e. if we introduce some debouncing to writing to the URL) - this opens a can of workms, in terms that the final state will not be predictable and the user won't be sure what changes did persist and which hasn't. Going Back will **not be deterministic**.

* "constant serialization" means:

- in auto mode: upon user touching any of the params
- in normal mode: upon user clicking the Run button

src/algorithms/types/Param.types.ts
dkozar: It might be irrelevant if the user custoized some params. But in general, it isn't.

I think `curren`t should still be send over the wire (via URL). Because, if i want to select one of the existing, default scenarios (say "Croatia") and I want to send the link to someone, they should see "Croatia" selected in the dropdown.

Basically, we need to persist all **the state needed to reconstruct the current screen**. Without persisting `current` you can't do that.

jsmith: Hmm it might not be irrelevant to the user who is sharing their parameters (ie. when they don't modify any params like you said) but the value is irrelevant upon deserialization. It's true that removing `current` will make it so we can't reconstruct an identical screen but doing so is not technically correct (the state is being initialized from the URL parameters, not the scenario defaults).

ivan-aksamentov:
@dkozar Thanks Danko. The restructuring is indeed useful.
Even if there are sync problems, I think we cannot make it even more broken that it is now. Merging.

src/algorithms/types/Param.types.ts
dkozar: Then we should really consider not to have the scenario as a dropdown.

We should really have a "Load existing scenario" button, which opens a popup window with a number of scenarios to load: https://github.com/neherlab/covid19_scenarios/issues/397

PR #388, opened 2020-04-03
Feat/invert mitigation

rneher:
## Description
This PR changes the mitigation intervals from "relative transmission" to "reduction". That is a 20% reduction would be specified as 0.2 rather than 0.8. This also adds labels to the plot.

The python scenario generation, the conversion form intervals to time series. 

## Testing
not expected to break anything.


rneher:
We could change the values to % in line with most other numbers in the app: 0.2 -> 20%

PR #390, opened 2020-04-03
Introduce JSON schema into the app

ivan-aksamentov:
## Related issues and PRs

 - Contributes to https://github.com/neherlab/covid19_scenarios/issues/283

## Description

This attempts to integrate JSON schema into Typescript side of the application: types, serialization, validation of inputs and outputs of the application.

Types have to be reusable across the parts of the application, and across application-data boundary as much as possible.

Our primary targets for typing, serialization and validation are:
 - [ ] ***IMPORTANT!!!*** URL parameters in/out (this feature has to be released in a couple of days at most)

Types, serialization, validation. Which means pretty much everything.

## Testing

To come, hopefully


ivan-aksamentov:
Currently implemented:
 - [ ] schemas
 - [ ] generate types from schemas
 - [ ] run generation as a build step

 Immediate TODO:
 - [ ] scaffold schemas necessary for URL params
 - [ ] validate URL inputs/outputs with AJV
 - [ ] (de)serialize URL inputs/outputs with ?

TODO:
 - [ ] run generation and build on schema changes (adjust nodemon config)
 - [ ] ...

ivan-aksamentov:
Switched to quicktype and added type generation for Python.

Generate TS types:
`yarn schema:totypes`
(also a part of `yarn dev`, so no need to worry most of the times)
Types will appear in `src/.generated/types`.

Generate Python classes:
`cd data && make types`
Python calsses will appear in `data/generated_types` and can be imported from there.

The generator script is here:
https://github.com/neherlab/covid19_scenarios/blob/1c453b038572c9250bbf0bd6c1972ebaf265f3f9/tools/convertSchemasToTypes.ts

As a pleasant addition, quicktype also generates serializers/validators - in Python from/to `dict`, in TS, for some reason from/to string JSON, not object. But we are not forced to use those. For TS there is a "cannonical" `ajv` and for python  - `jsonschema`.

Neither types, nor serializers are currently used anywhere, they are just sitting there.

Also, schemas have some types not specified. Age distribution in python is a bit ugly. Schema needs to be massaged a bit.

noleti:
I have somequestions:
- how do you generate the schemas currently? By hand? 
- It seems that the 'time' string is not really validated, apart from checking that it is a string, correct? Is it possible to enforce a format for it?
- The data/generated_types/ scripts are to parse strings into json, but we are currently going via .tsv into json. So we would run those validation scripts after generating the json, e.g. as part of `generate_data --output-scenarios foo.json`?

ivan-aksamentov:
@noleti 

> * how do you generate the schemas currently? By hand?

If by "currently" you mean "before this PR and before JSON schema introduction", "on master", then there were no schemas basically. Nick just wrote a bunch of classes manually, that "correspond" to typescript types in the app. Of course the "correspondedness" had to be maintained by hand.

If by "currently" you mean "currently in this PR's branch", then @thoo1  wrote the `.schema.json` files by hand and added a test with ajv in https://github.com/neherlab/covid19_scenarios/pull/297, and then Nick has refined schemas a little bit. 

There are command-line tools for converting typescript types into JSON schemas. But it's not such a big of the deal, the format is quite concise and intuitive: https://json-schema.org/understanding-json-schema/reference/

One thing is, that I've mostly seen schemas written in YAML format these days, not in JSON, so that it's  a bit prettier and can have comments. Again, this can be converted automatically.

> * It seems that the 'time' string is not really validated, apart from checking that it is a string, correct? Is it possible to enforce a format for it?

Haven't checked our schemas much yet, they are definitely suboptimal at places.

Yes, there are ways to specify the string format: 
https://json-schema.org/understanding-json-schema/reference/string.html
In the most generic case - with regex, but there are also some pre-defined formats. Then, this format will be enforced by the validator during runtime or, if language is capable, also in type (for example, `Date` type will be used in Typescript if `date-time" is specified as schema type)

> * The data/generated_types/ scripts are to parse strings into json, but we are currently going via .tsv into json. So we would run those validation scripts after generating the json, e.g. as part of `generate_data --output-scenarios foo.json`?

Ideally, you run validation right before you read or write from/to the external output/input. In this case you write a string into the file. So before passing string to `f.write()`, just pass it through the validator first. Some validators also act as serializers, which means that they can not only validate a string, but also produce a correct string from an object (and deserializers are for the opposite operation).

Using serializers generated by `quicktype` is optional, I am not certain on how they useful and practical. The comment on top of files says:

```
To use this code, make sure you
    import json
and then, to convert JSON from a string, do
    result = case_counts_from_dict(json.loads(json_string))

```
This example uses `json.loads()` as a deserializer (string -> object) and `case_counts_from_dict` as a validator.

But you can also validate the dict without deserialization:
```
result = case_counts_from_dict(case_counts)
```

Or just use the provided class as it is done now with manual classes. But in this case, we only get the type. The validation and serialization should be done separately. Package [jsonschema](https://github.com/Julian/jsonschema) is one of the popular ones:

```
from jsonschema import validate

validate(instance=case_counts_object, schema=schema)
```

ivan-aksamentov:
@noleti @rneher @nnoll 
In general, do you think this solution would work for data folks?

The disadvantage I see are:
 - there is an additional build step - mitigated by Makefile
 - requires Node.js and yarn - okay, we can do without yarn, but quicktype is written in TS.

noleti:
Thanks for the detailed reply. Yes, I was referring to the 'currently in this PR's branch', in case I would need to change something (e.g., source strings in scenarios, if we eventually want to support that). 

I think the python classes alone are probably already very useful for the `data/` things, as so far there are just dicts of lists of dicts being used... One would ideally rewrite the parsers and other tools to use the classes, but that should hopefully be easy. 

Regarding .tsv parsing: we would have to adapt the current .tsv parsers to write in the right class. I assume it does not make sense to validate afterwards in the same code? We should know that the class is of right type at that point, correct? I assume we would validate() when parsing .json files, for example, correct?

ivan-aksamentov:
@noleti Classes by themselves do not validate anything, so the advantages of converting existing non-class code to classes is quite slim. My understanding was that the existing generator already uses the classes we need.  And runtime validation is orthogonal to classes - you can validate and serialize any object or string, whether instantiated with a class or not.

All I care about right now, is that the final data produced is compatible with the app at all times. That is the types are the same and values are valid on both ends.

We will definitely validate on the app's end when importing, but it wouldn't hurt if the final data script validates what it outputs. This would be useful for dev workflow (fail early) and for automating the updates.

I am not sure about the intricacies of .tsv to .json conversion part and how it works, so for now I leave this decision up to data maintainers. Technically, because it is internal to the "data repo", you folks can do any kind of circus you'd like there, as long as the app receives the correct data :) So JSON schema will probably be an overkill for the internal types (unless they intersect with the output types).

However, the final output has to be validated and/or serialized using schema-aware validator/serializer. Quick solution is to just take the whatever string you write to the final .json file, and validate it as it is. That's all that is needed right now.

In general it is good practice to use proper validation and serialization every time the data leaves or enters system memory - whether to disk, network, browser local storage etc., because these are the "untrusted" sources. So it might make sense to also implement a schema for parser-to-generator boundary (raw data -> intermediate .tsv), but that is out of my scope right now.

noleti:
AFAIK we don't use classes for anything yet in the parsers, .tsv or json generation. The only datastructures we use are essentially the same as your classes now, from what I can tell. I will wait for this PR to be merged, and then I will make a PR to introduce the classes (+ validation) at least for the generation of .json that the app will consume.

noleti:
I'm trying to test this now, and I have some issues. In particular, I am using `jsonschema` as suggested. My MWE is as follows (run in the `data/` directory with `python3 test.py`). 

```
from generated_types.CaseCounts import case_counts_from_dict
from jsonschema import validate
from json import load

if __name__ == "__main__":
    test = {'Germany': [{'time': '2020-03-20', 'cases': 1, 'deaths':1}]}
    case_counts = case_counts_from_dict(test)
    print('Generated case_counts')
    with open('../schemas/CaseCounts.schema.json', 'r') as fh:        
        schema = load(fh)
    
    validate(instance=case_counts, schema=schema)
```

Output is as follows (shortened):

> $ python3 test.py 
> Generated case_counts
> Traceback (most recent call last):
>   File "test.py", line 12, in <module>
>     validate(instance=case_counts, schema=schema)
>   File "/usr/local/lib/python3.6/dist-packages/jsonschema/validators.py", line 934, in validate
>     raise error
> jsonschema.exceptions.ValidationError: <generated_types.CaseCounts.TimeSeries object at 0x7fbcd6d0a3c8> is not of type 'object'
> 
> Failed validating 'type' in schema['additionalProperties']['items']:
> [...]
> On instance['Germany'][0]:
>     <generated_types.CaseCounts.TimeSeries object at 0x7fbcd6d0a3c8>

Am I doing something wrong?

nnoll:
I agree with everything @noleti states. Adding targets to a Makefile is easy enough. However, I'm going to hold off working on the data end until this PR gets merged and the data serialization gets sorted out. Otherwise I'm coding against a moving target.

My thoughts on the places this will touch. There are a few places where classes are used in the current approach. The most important is within scripts/scenarios.py which as @ivan-aksamentov states I hand coded Python classes to match our input parameters one-to-one. This isn't a requirement, but it sure makes the output JSON an easy step! The first order of business would be to move this to an import statement from a generated file as @ivan-aksamentov suggests. Second, as @noleti was suggesting, this could be in the case_counts.json generation. This data format should never change though and is not an external boundary to the app so I static schema files could work here...

noleti:
I would assume that `data/types/`will be in the repo, then data-only contributors to parsers will never have to touch/know any of this. If we introduce the CaseCounts class in the parsers, they should also be easy enough to use (exactly like the previous datastructure, I assume). So I don't think this would be an issue for new contributors as long as the class definition is in the repo.

nnoll:
Yes that would be ideal. A parser script would just import this class and not have to worry about generating it or where it comes from. I'm on board!

noleti:
I found the issue in my validation MWE now: `validate(instance=case_counts_to_dict(case_counts), schema=schema)` is required. I might still miss something here?

ivan-aksamentov:
Updates:

 - converted schemas to YAML and cleaned them up, so that you can actually see the types clearly


Let's not:
 - add documentation fields yet (title, description etc.), until we establish proper types
 - define types as narrowly as possible. Read ["Understanding JSON Schema"](https://json-schema.org/understanding-json-schema) and unleash your type creativity.
 - defined ids for the time being
 - keep types as reusable as possible. For that we may need to split "definitions" into multiple files and then refer to those from other files (like import in Python or JS)


@noleti The usage now should be something like this:
```
from datetime import datetime

import yaml
from jsonschema import validate

from generated_types.CaseCounts import case_counts_to_dict, EmpiricalDatum


if __name__ == "__main__":
    case_counts = {
        "Germany": [
            # Usage of classes is encouraged for build-time type checking
            # and better code completion, but not mandatory. Could be a dict.
            EmpiricalDatum(
                cases=1,
                deaths=1,
                hospitalized=1,
                icu=1,
                recovered=1,
                time=datetime(2020, 3, 20, 0, 0),
            )
        ]
    }

    with open("../schemas/CaseCounts.yml", "r") as f:
        schema = yaml.load(f, Loader=yaml.FullLoader)
        validate(case_counts_to_dict(case_counts), schema)

```

 `jsonschema` does not understand the classes generated by `quicktype`, it can only operate on dicts. `quicktype` wrappers can convert dict <-- > class object, so the conversion is necessary. We may define a set of function to wrap this logic.


I would encourage usage of classes explicitly when creating objects (like `EmpiricalDatum`), instead of using the generated `*_from_dict` functions, as this would allow for better dev experience.


Commiting generated files to source control defeats the purpose of schemas as a single source of truth, as now there will be 3 sources (instead of 2 before schemas). We would rely on people updating the generated files on schema updates, and people are typically unreliable creatures.


Regarding the dependency on Node, we may try to use a Python generator, but the available ones look much inferior compared to `quicktype`. Also, how do you imagine someone contributing to data output without running the app afterwards? (or are we doing Knuth style "Beware of bugs in the above code; I have only proved it correct, not tried it"? :D ). Schemas should not affect parser writers, I think, unless we are doing something terribly wrong there...

noleti:
Thanks for the detailed comments again. I will wait for this to be merged before implementing changes in `data/`. The merge should not break anything there anyways.

I would still prefer to have the generated python types in the repo, as I said I think we could put a README.md in the folder to discourage people from editing directly. Personally, I only installed `yarn` now for this now, as I am not the biggest fan of .js, nodejs and their packet managers etc (bad experiences in the past). I agree  with @nnoll that we should keep the entry barrier for people contributing parsers low. They can check the output in the `case-counts/` folder, and don't need the app locally (or any .js related stuff). This is how I contributed in the `covid19_scenarios_data` project in the last weeks.

ivan-aksamentov:
We did a pair coding session with @brunorzn 
Updates: 

 - we had to change 2 of the schemas:

   - Scenarios is now an array of objects with a "country" key each, instead of an object where country names are the keys.

       schema: https://github.com/neherlab/covid19_scenarios/blob/3b8b29991c1e571a233e78369619a34180d8d609/schemas/Scenarios.yml
       example: https://github.com/neherlab/covid19_scenarios/blob/3b8b29991c1e571a233e78369619a34180d8d609/src/assets/data/scenarios/scenarios.json

   - Same for case counts. 
        schema: https://github.com/neherlab/covid19_scenarios/blob/3b8b29991c1e571a233e78369619a34180d8d609/schemas/CaseCounts.yml
        example: https://github.com/neherlab/covid19_scenarios/blob/3b8b29991c1e571a233e78369619a34180d8d609/src/assets/data/case_counts.json

Could someone jump in and adjust data for this new layout? cc @nnoll @rneher @noleti 

 - Input types in the app are mostly replaced with generated ones

 - Implemented validation and serialization for case counts and scenario data
       TODO:
           - [ ] add unit tests for validation/serialization functionality (to make sure they break on wrong input and to find places where the types are too relaxed)

 - Field "ICU"  in case counts was renamed to "icu" because schema tools assume using coding conventions, which we should
     TODOs: 
            - [ ] adjust both the app and data for the lowercase "icu" field
            - [ ] also verify other misnamed fields, like "AllParams.population.ICUBeds"

brunorzn:
I've just realised that countryAgeDistribution should follow the same "array" pattern (same as scenario, cases count).
Working on it now.

noleti:
I tested this now for CaseCounts and Scenarios, and pushed my changes to this PR. Two issues:

- FullLoader requires a PyYAML from March or so, which lead to me having to update yaml (via `pip3 install --upgrade PyYAML`). BaseLoader failed on the multipleOf with some cryptic error (1 is not a number). I see we have a Pipenv.lock, but I tried to get pipenv to work now for 20 minutes and it looks like this leads to me having to update to python 3.8, install user repos on by ubuntu, ... the usual webapp install depency bleeding edge mess. Is it possible to not use FullLoader?
- 'cases' cannot be None at the moment, but I believe this is tolerated in the app. If it is not acceptable, we will need logic in `data/` to remove such entries from the data before saving to .json.

Validation of cases currently fails for the reason state above.

rneher:
I added `None` as allowed type to cases. this happens for example for cases in france where only deaths are reported in some places. Also fixed a problem in export of case counts.

rneher:
validation is working for me in python 3.6.8 and yaml version 5.1

ivan-aksamentov:
@noleti 

> FullLoader requires a PyYAML from March or so

Sorry for the mess, I just copied the first snippet I found on the web ;)
All we need is to load this yaml, no matter how. And this is a relatively standard yaml too. If you know an alternative library, feel free to push.

noleti:
Thanks, lets stay with FullLoader/ PyYaml. It all works for me now, anyways. From my side, we can merge after fixing the failing tests. The failing test seems to be due `src/algorithms/validateSchema.test.ts` still using .json schemas (instead of .yml)

ivan-aksamentov:
@noleti Thanks for testing all that! Yes, there is a little cleanup needed before merge. Will do this now.

Issue #385, opened 2020-04-03
Improve Landing page redirect logic


ivan-aksamentov:
# 🙋 Feature Request

In short: use `connected -react-router`.

## 😯 Describe the feature

 - [ ] The `<Redirect/>` from `react-router` directly circumvents the `connected -react-router`. I think that breaks the back/forward navigation in some cases, because all other pages use `connected -react-router`. I would prefer to move this out of the component and `dispatch(push(url))` where `push` is the action in `connected -react-router`.
- [ ] make sure URL serialization/deserialization still works

## Related

 - https://github.com/neherlab/covid19_scenarios/pull/163
 - https://github.com/neherlab/covid19_scenarios/issues/107

PR #392, opened 2020-04-04
fix Improve Landing page redirect logic #385

guandiyiyi:
## Related issues and PRs
fix Improve Landing page redirect logic
Contribute to #385
 - 

## Description

- Fix the reloading issue in lading page -- when reload the scenario page, make sure it doesn't redirect to landing page
- fix browser back/forward issue

App.tsx:
change to use history.push for redirection to landing page

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1,8 +1,11 @@
 import React from 'react'
-
 import { useSelector } from 'react-redux'
+import { useHistory } from 'react-router-dom'
 import { withErrorBoundary } from 'react-error-boundary'
-import { Switch, Route, Redirect } from 'react-router'
+import { Switch, Route } from 'react-router'
+
+import _ from 'lodash'
+
 import ErrorPage from '../pages/ErrorPage'
 
 import Layout from './Layout/Layout'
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -14,10 +17,13 @@ import { State } from '../state/reducer'
 function App() {
   const shouldGoToLandingPage = useSelector(({ ui }: State) => !ui.skipLandingPage)
   const initialQueryString = window.location.search
-
+  let history = useHistory()
   return (
     <Switch>
-      {shouldGoToLandingPage && <Redirect exact from="/" to="/start" />}
+      {shouldGoToLandingPage &&
```
ashwinGokhale, line 23: I think we can refactor to the use "to" prop in the Redirect component rather than invoking a function which causes a side effect inside the rendering of this component. See react router documentation: https://reacttraining.com/react-router/core/api/Redirect/to-object

guandiyiyi: Yup, make sense. I think the original one doesn't work is because it ignores the search, will give it a try tonight!

ivan-aksamentov:
@guandiyiyi This is not what the linked issue asks for! It asks to remove the current incorrect implementation and replace it with something else, not to patch it. Please read the checkbox items in the issue, one of them suggests a definitive fix.

This logic should go before any React component, possibly before router initialization.
`connected-react-router`, it's the router we use in the app. Not `react-router`. We will not be reinventing our own routing solution either. I will not accept anything related to the raw history API.

Let me know if you have any questions. If you have an alternative opinion, please manifest it early. :)

Issue #348, opened 2020-03-23
Keep track of sources for populationData.tsv


nnoll:
As of now we don't track where we get this data from. I propose we amend sources.json to have two top-level fields: ["case-counts", "populationData"] in which we store all information from which we derive our data. 

The data below "case-counts" would be equivalent to what's now, while "populationData" would hold analogous level information for hospital bed/ICU capacity. Keys in the JSON should be equivalent to the country name.

noleti:
@nnoll I will start working on this over this weekend. Any updates wrt latest changes?

PR #394, opened 2020-04-04
Support for sources of data in populationData.tsv

noleti:
## Related issues and PRs

 - Fixes https://github.com/neherlab/covid19_scenarios/issues/348

## Description

Provides a way to record the sources of total population, hospital beds, and ICU beds in populationData.tsv

- data/populationData.tsv
- data/scripts/population.py
- data/scripts/scenarios.py
- data/static_population_data.tsv (seems to be outdated file that is unused, this PR removes it)

## Testing

- python3 data/generate_data.py --output-population foo.json
- python3 data/generate_data.py --output-scenarios foo.json

Both scripts should run. Population.py will not actually change anything. The generated scenarios.json will now contain a top level key with a dict of sources.

As many entries in populationData.tsv were directly added by @rneher , I put 'None' as sources for now. I think the scripts expect some value, so just leaving the entry empty might cause problems. This might be improved later.


rneher:
Thanks @noleti. Very useful to keep track of the sources for these numbers and remove the obsolete columns. Just one quick worry. The `AllParams` object now has a key that is not reflected in the types of the app. I would therefore hold off adding it to the exported json only we decide whether we need this info in the app and how to store it..

noleti:
Removed the 'sources' key from json output now. In my mind, we eventually could expose the string to the user in the app, e.g. as part of the 'more information' button already present next to each parameter. So it should be ideally human readable.

rneher:
I agree with exposing this info eventually to the user in the app (just trying to limit the number of moving parts ;) ).

Issue #106, opened 2020-03-22
Optimize for search engines


ivan-aksamentov:
We want people to be able to find out about our app in different ways, including by searching in common search engines. It would be nice to have at least basic Search Engine Optimization (SEO).

What has to be done here?

Related
#105

ivan-aksamentov:
This has been addressed in issues above. We will open more focused issues if further improvements are necessary.

PR #396, opened 2020-04-04
Added metadata to index.ejs to improve seo

### LINKED ISSUE #105: Optimize for social media
As we rely on social media to increase the outreach, it would be nice to: 

  - [ ] double check title tag and favicons for different devices (see `static/` directory)
  - [ ] add social media meta tags for common western social media, like Twitter, Facebook, Google
  - [ ] add social media meta tags for common social media in asian-pacific region (I am not familiar with these unfortunately)
  - [ ] ensure that the sharing on various social platforms produces correct and visually appealing links and previews, with a proper title and thumbnails

Feel free to add your features and suggestions.

Related:
 - #106 
 - https://github.com/nfl/react-helmet
 - https://github.com/staylor/react-helmet-async

Please commit and PR often, to make sure that the merges are small and there is no work wasted on different people doing the same thing. Discuss in the chat.

ashwinGokhale:
## Related issues and PRs

 - Fixes #105 and #106 

## Description

Added meta tags to optimize for SEO and social media

Note:
* Removed extraneous icon files. Browsers (both mobile and desktop) can automatically optimize for different file dimensions, so no need to bloat the output bundle when we don't need to.
* Since we are not server side rendering our pages, we cannot use `react-helmet`. If we decide to go down the SSR route, I would suggest `Next.js`.

index.ejs
static/

## Testing

Ran lighthouse report in Google Chrome:

image


ivan-aksamentov:
@ashwinGokhale Thanks!

`react-helmet` has nothing to do with SSR, other than it does support it.

It/s primary purpose is to change tags in the `<head/>` dynamically. All you need to do is to move all the stuff you just added to the head of an HTML file, as children of `<Helmet/>` component in a JSX file somewhere in the app's compoenent tree.

```
            <Helmet>
                <meta charSet="utf-8" />
                <title>My Title</title>
                <link rel="canonical" href="http://mysite.com/example" />
            </Helmet>

```

This way we can for example change the tab title when user selects a country, or change the twitter card image or canonical url depending on some conditions, etc.

Not essential and not a priority right now, but why not.

PR #399, opened 2020-04-04
Feat/Firebase sync WIP

miksa1987:
## Related issues and PRs

[#222](https://github.com/neherlab/covid19_scenarios/issues/222): Cloud based syncing/backup of scenarios/user preferences 
 

## Description

As discussed on the issue, I've been doing some work to get the app to sync settings and scenarios with Firebase. For now, it will sync preferences, but it does have functions for handling scenario data. For setup, this would need Firebase config variable which is available from their settings, placed into src/firebaseconf.ts. See src/firebaseconf.example.ts.

So, there's still some things I'd like to implement in here before I could say this is ready for merging. 
I'm having varying amounts of time every day, so this might take a while. Here's a list of what I'm thinking at least:

- [ ] Different login methods (Google, Facebook, Twitter ... ) Firebase has plenty of these.
- [ ] Icon for user menu. Right now it's pretty ugly just saying "USER".
- [ ] Somehow indicate if a user is logged in or not on the menu icon. What about a user icon with a question mark if not logged in, and a social icon perhaps for logged in?
- [ ] User can download their data.
- [ ] User can delete their data.
- [ ] Fix the language toggle inside context menu. Maybe a submenu? Or a popup?

@ivan-aksamentov @rcbevans @gj262 what do you guys think? I could use your comments on these. Right now this is still a pretty rough mockup.

## Testing

TBD


ivan-aksamentov:
Hi @miksa1987 , sorry for the delay.

We, as maintainers, never really thought about this kind of storage and I think maintaining correctness of the database would be particularly hard given that our data schemas change so quickly. We decided against this feature.

But thanks for your work anyways!

PR #403, opened 2020-04-04
Improve rendering performance

### LINKED ISSUE #382: [IMPORTANT] Fix delay when adding mitigation intervals
# 🐛 Bug Report

## How to reproduce

Steps to reproduce the issue:

 2. Click "Add" button in "Mitigation"
 3. Observe that it takes a while for the new interval to appear in the table

## 😯 Current Behavior

New intervals appear after delay

## 🤔 Expected Behavior

New intervals should appear instantly

## 🔦 Context

It's either "Add" button is wrongly debounced or it waits for the algorithm to finish, or both.

Importance:  Intervals is a commonly asked for feature. Should be included into the next release.

## 🌍 Your Environment

Any

## Related

 - https://github.com/neherlab/covid19_scenarios/pull/323
 - https://github.com/neherlab/covid19_scenarios/issues/14

gj262:
## Related issues and PRs

Fixes #382 

## Description

Improve app rendering performance so that the act of adding a mitigation line is responsive.

Prior to this change, updates from the form were debounced for a second before changing scenario state. This was due to updates causing re-renders of the result graphs.

This change:

- Reduces the debounce interval to 100ms
- Disables all Rechart animation
- React.Memoizes components so they will no longer re-render just because their parent component re-renders. They will only re-render when a prop changes.
- Removes unused props, variables.

Adding a mitigation is still a little laggy due to the fact that the new mitigation causes a render of the expensive DeterministicLinePlot.

Performance in general and adding a mitigation line in particular.

## Testing

With the results rendered:

Add a mitigation. 
Delete a mitigation.
Tweak parameters.
Repeat with autorun.


ivan-aksamentov:
@gj262 This indeed makes things much snappier, but does not remove the delay on "Add", just makes it shorter.

Also, my worry is the `React.memo`'s shallow comparison. We may have to customize this with a deeper comparison, in case if someone somewhere modifies object's internals without updating the whole object. We need to watch out for the weird memoization-related-bugs.

Overall I am eager to merge, but will definitely keep the issue open.

@rneher @nnoll @brunorzn @dkozar  What do you think? It's a huge usability improvement, especially on slow machines, but there is a slight risk of bugs: component may not re-render if props change incorrectly (which I am almost sure is happening somewhere). Do we need to merge this before the release?

gj262:
> @gj262 This indeed makes things much snappier, but does not remove the delay on "Add", just makes it shorter.

Yeah @ivan-aksamentov, there would need to be separation between state that is used to control the form and the state for the `<DeterministicLinePlot />`. i.e. another scenarioStates. So the form could add the mitigation and have it render and the graph would render later. I talked about one possible approach to this in #382 before I headed down the React.memo route. (This assumes that it is not possible to improve the rendering performance of `<DeterministicLinePlot />` which is reasonable given its graphiness).

> Also, my worry is the `React.memo`'s shallow comparison. We may have to customize this with a deeper comparison, in case if someone somewhere modifies object's internals without updating the whole object. We need to watch out for the weird memoization-related-bugs.

A very valid concern. There is deep cloning in the reducer but it is hard to be confident that other state is not being subtly mutated e.g. severity table. On the other hand, I think I introduced React.memo around the ResultsCard in an earlier fix so the weird memoization-related-bugs may already be there.

rneher:
It definitely is much snappier. I see the concern. But keeping it in mind fill hopefully make us look in the right place when things act up. I am in favor or merging.

nnoll:
I'm all for speed/low latency. I approve.

gj262:
I'm going to close this in favor of #437 which is simpler, cleaner, neater :)

Issue #119, opened 2020-03-22
Add more e2e tests


ivan-aksamentov:
# 🙋 Feature Request

We need to cover some more user interactions.

Basically, any usage scenario is:
 - change values on the right
 - observe results appearing on the left
 - [optionally] change values again
 - [optionally] observe results changing on the left

Making sure that we at least don't crash and render something on simple interactions, like setting wrong values in the number boxes would be a great start.

The central part of the app are plots and tables in the results card. How could we test them?

Testing other pages probably not necessary, they are almost plain text.

maktarsis:
@ivan-aksamentov can you clarify what do these terms mean "change values on the right" and "observe on the left"

I mean like here in the screenshot, only in reverse order, right?
https://prnt.sc/rohsip

ivan-aksamentov:
@maktarsis Right. Mixed up the two.

PR #406, opened 2020-04-05
test: add tests for scenarios

Joluma:
## Related issues and PRs
Contributes to #119  

## Description
Adding e2e tests for the scenario selector
- select a different scenario
- the correct values are set in the population card
- the correct values are set in the epidemiology card

Tests

## Testing
run `yarn e2e` for a full test run
run `yarn e2e -s "**/scenario*"` to only run the scenario tests

scenariospec

### DIFF
--- a/cypress/integration/scenarios.spec.ts
+++ b/cypress/integration/scenarios.spec.ts
@@ -45,6 +45,8 @@ context('Language switcher', () => {
   })
 
   for (const scenarioKey of firstTenScenarios) {
+    const { population } = scenarios[scenarioKey];
+
     describe(`Switching to "${scenarioKey}"`, () => {
       it(`should change the text on selector to "${scenarioKey}" correctly`, () => {
         cy.get('@ScenarioDropdown')
@@ -56,6 +58,16 @@ context('Language switcher', () => {
         cy.get('@ScenarioDropdown')
           .find('[class$="singleValue"]')
           .should('have.text', scenarioKey)
+      })
+
+      it('should change the values for the "Population" card', () => {
+        cy.get('.card--population').as('PopulationCard')
+
+        for (const key in population) {
+          cy.get('@PopulationCard')
+            .find(`[name="population.${key}"]`)
+            .should('have.value', String(population[key]))
+        }
       })
     })
   }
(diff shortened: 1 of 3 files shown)


cypress/integration/scenarios.spec.ts
```diff
@@ -0,0 +1,83 @@
+/// <reference types="cypress" />
+
+import scenarios from '../../src/assets/data/scenarios/scenarios.json'
+
+const scenariosKeys = Object.keys(scenarios);
+// testing all the scenarios would be extremely long
+const firstTenScenarios = scenariosKeys.slice(0, 10);
```
Joluma, line 7: There're **237** scenarios, running test on all of them will be time consuming, but if it' necessary we can remove this line and directly use `scenariosKeys` on line 49.

cypress/plugins/index.js
```diff
@@ -12,7 +12,7 @@
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
 
-import webpack from '@cypress/webpack-preprocessor'
+const webpack = require('@cypress/webpack-preprocessor')
```
Joluma, line 15: This line was changed recently, can somebody explain why? We can't use import in this file for now without adding some webpack loader configuration.

noleti: Well the commit  https://github.com/neherlab/covid19_scenarios/commit/1ee53fcdda989967b9a6c9175953efafac5d25d1 was by @abrie  with the comment ' Change 'require' to 'import' to fix eslint no-var-requires.'

Joluma: Thanks @noleti.
I checked that PR but I don't know the reason, was this line breaking the build or something? Maybe @abrie can explain which part I need to change (worst case I'll just add a //eslint ignore).

noleti: I assume it was just eslint complaining: https://palantir.github.io/tslint/rules/no-var-requires/

abrie: @Joluma indeed it _was_ breaking the build because of a hard limit to eslint errors. But the fix breaks the 2e2 tests, which I don't think anybody has been running (until now). I agree with your solution to explicitly ignore the error.

PR #408, opened 2020-04-05
fixed typo and grammar mistakes

TeepaBlue:
## Related issues and PRs

 - 

## Description
I fixed typo and grammar mistakes.


src/assets/text/updates.md
```diff
@@ -25,15 +25,15 @@ A smaller R0 implies that infection control measures are more effective, so this
 Our previous code stepped the ODE forward in 6h intervals. This large step size
 results in inaccuracies, in particular as we are using first order integration scheme.
 This step size is now reduced to 0.05 days. This can change the output of the model
-by a few percent. We will soon move to a higher order integration scheme for 
+by a few percents. We will soon move to a higher order integration scheme for
```
noleti, line 28: I would rather say `slightly` instead of 'by a few percent'. `percents` is wrong in my opinion, as it is always per 100. I'm not a native speaker, though.

noleti: Thanks for the PR. Everything looks good except this one comment I have. What do you think?

rneher:
I agree with @noleti, other than for the `percents` change this looks good.

src/assets/text/updates.md
TeepaBlue: Yes, it's correct. I'm sorry for that.

rneher: Thanks!

Issue #409, opened 2020-04-05
Simulation.timeRange doesn't serialize correctly


rneher:
##🐛 Bug Report
When running in a new tab or just reloading, the simulation doesn't actually run because the time range is set to 0--0. This is what sits in the URL:
`simulation~(simulationTimeRange~(tMin~0~tMax~0)~numberStochasticRuns~0)`

## How to reproduce
 1. Open the application in a browser
 2. run, followed by run in new tab

## 😯 Current Behavior
The plot starts at 1970, no simulation results are output.

## 🤔 Expected Behavior
the timeRange should be serialized correctly.

## 🌍 Your Environment
Linux 16.04 Firefox and Chrome.

PR #410, opened 2020-04-05
Fix/date serialization

rneher:
## Related issues and PRs

#409 

## Description
This adds a simple (brittle) conversion from the simulation.timeRange date  strings '%Y-%m-%d' to a  date object and timestamp. This is most likely not how this should work. But it fixes the bug. 

conversion of the state to object that goes into serialization.

### DIFF
--- a/src/components/Main/state/serialization/DtoConverter.ts
+++ b/src/components/Main/state/serialization/DtoConverter.ts
@@ -22,9 +22,7 @@ import { MitigationInterval } from '../../../../algorithms/types/Param.types'
  */
 const getTime = (potentialDate: Date | string): number => {
   if (typeof potentialDate === 'string') {
-    try {
-      return new Date(potentialDate).getTime()
-    } catch (error) {}
+    return new Date(potentialDate).getTime()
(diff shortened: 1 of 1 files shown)


rneher:
@dkozar -- there is likely a reason why this isn't the road you chose. I was largely curious why this happened and wondered whether I could fix it...

rneher:
I saw the other discussion -- feel free to disregard...

dkozar:
The readon was that I assumed this is a bug, which needs to be fixed at the source. Because vetting the string instead of date looks like an error.

So I wrote this as a throwaway code, with a goal to temporarily mitigate the problem (so we can merge my PR) until we fix at at the source. Your change is good, but it masks the problem to the extent that it stops being the problem - and we’ll probably never address it at the source. :)

rneher:
yeah, I just papered the problem over. We definitely need to look into the fact why dates are pulled out differently from different date pickers.

gj262:
Looks good to me. Those string dates are coming from scenarios.json and need to be handled.

PR #402, opened 2020-04-04
#382/Experimental fix for UI lag when adding/removing mitigation intervals.

abrie:
## Related issues and PRs

#382 

## Description

This PR tries to fix the UI lag when adding a new mitigation interval. It removes the debounce delay in `setScenarioCustom()`. The reason for debounce isn't immediately obvious, I hypothesize that it's unnecessary, and am drafting this PR for feedback/testing.

Looks like the primary source of delay is the Deterministic line plot. Wrapping it in a `React.memo()` greatly improves the lag.

## Impacted Areas in the application

- Mitigation interval UI
- Possibly the DeterministicLinePlot component, because it is now wrapped with `React.memo()`

## Testing

Start a dev server, load the app, scroll to the Mitigation section, click "Add" to add a new interval, click the trash icon to remove it:

add-mitigation-interval

PR #378, opened 2020-04-03
Drag/Resize Scenario and Results vs. Zoom in/out chart

guandiyiyi:
## Related issues and PRs
"Expand" the results panel
#284
 - 

## Description
Allow drag and resize the left and right panel
working in progress to allow zoom in/out

## Impacted Areas in the application
Main page: Scenario card and Results card

PR #412, opened 2020-04-05
#382/Fix UI lag when modifying scenario fields. 

### LINKED ISSUE #382: [IMPORTANT] Fix delay when adding mitigation intervals

abrie:
## Related issues and PRs

Fixes #382, similar to #402, #403 
 
## Description

This PR attempts to reduces UI lag when editing Scenario parameters. The lag is primarily due to the heavy `DeterministicLinePlot` component which has several complex state dependencies. The current codebase mitigates the lag by adding a 1000msec debounce to the `setCustomScenario()`. That works – except it introduces a new lag (of 1000ms) to the mitigation interval editor. #402 and #403 fix the problem by wrapping the DeterministicLinePlot with `React.memo()`, but this may have unintended consequences because of the complex state.

This PR removes the debounce from `setCustomScenario`, and moves it to ~~the render method of~~ a wrapper component around  `DeterministicLinePlot`. The net effect is the same, and we avoid the potentially troublesome memoization.

~~Drawback is that the DeterministicLinePlot is re-rendered every 1000ms; but if there is no change then the cost is minimal.~~ (no longer an issue)

~~**Another drawback is that the graph interactions are now sluggish**~~ (no longer an issue)

- General UI input responsiveness
- (re)rendering rate of Results Card, particularly DeterministicLinePlot.

## Testing

Test the fix by looking for UI sluggishness. The scenario inputs and mitigation editor should provide snappy feedback. Test with autorun enabled and disabled. Verify that tooltips are shown as expected when hovering over the charts.

add-mitigation-interval


src/components/Main/Main.tsx
```diff
@@ -163,7 +163,7 @@ function Main() {
       const mitigationIntervals = _.map(newParams.containment.mitigationIntervals, _.cloneDeep)
       scenarioDispatch(setContainmentData({ data: { mitigationIntervals } }))
     }
-  }, 1000)
+  }, 0)
```
abrie, line 166: Maybe just remove the debounce altogether...

src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -109,6 +111,9 @@ export function DeterministicLinePlot({
   const { t } = useTranslation()
   const chartRef = React.useRef(null)
   const [enabledPlots, setEnabledPlots] = useState(Object.values(DATA_POINTS))
+  const [rendered, setRendered] = useState(<></>)
```
abrie, line 114: We cache the render output as state. Initially it contains nothing.

src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -109,6 +111,9 @@ export function DeterministicLinePlot({
...
+
+  const [debouncedRender] = useDebouncedCallback(() => setRendered(render()), 1000)
```
abrie, line 116: Store the rendered output into the render state. The render function can take upwards of 60msec to finish, so we debounce to a maximum of once every 1000 msec.

src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -248,103 +253,111 @@ export function DeterministicLinePlot({
...
-
-                <Legend
-                  verticalAlign="top"
-                  formatter={(v, e) => legendFormatter(enabledPlots, v, e)}
-                  onClick={(e) => {
-                    const plots = enabledPlots.slice(0)
-                    enabledPlots.includes(e.dataKey) ? plots.splice(plots.indexOf(e.dataKey), 1) : plots.push(e.dataKey)
-                    setEnabledPlots(plots)
+  function render() {
+    return (
```
abrie, line 257: Here the original component's 'return' statement is wrapped into a function which will be debounced.

src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -248,103 +253,111 @@ export function DeterministicLinePlot({
...
+            )
+          }}
+        </ReactResizeDetector>
+      </div>
+    )
+  }
+
+  debouncedRender()
+
+  return rendered
```
abrie, line 362: Invoked the debounced render to (maybe) update the render state. Then return the (maybe updated) render state.

gj262:
This is kinda cool. I would never have thought of debouncing render(). It fixes the main culprit of the problem.

abrie:
> This is kinda cool. I would never have thought of debouncing render(). It fixes the main culprit of the problem.

There are some drawbacks; one being that the graph interaction is now sluggish. But perhaps there is a clean workaround. Worth a bit more work.

And, unfortunately this fix is not compatible with the the merge of #378.

gj262:
And the tooltips disappear if the mouse sits on the graph.

abrie:
Ok, got some fixes for these issues: https://github.com/neherlab/covid19_scenarios/pull/412#issuecomment-609518115 and https://github.com/neherlab/covid19_scenarios/pull/412#issuecomment-609517558.

The issues stem from the fact that there are two update scenarios for the component. Either its modified through props externally, or re-rendered from within through zoom/tooltip actions. The internal actions should not be debounced, and external actions must be debounced. So this fix moves the debouncing logic out of `DeterministicLinePlot` and into a wrapper class component (`DebouncedDeterministicLinePlot`). The wrapper takes care of debouncing changes to the props, while leaves internal events untouched.

src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -97,6 +97,24 @@ function legendFormatter(enabledPlots: string[], value: string, entry: any) {
...
+  shouldComponentUpdate() {
+    this.debouncedRender()
+    return false;
+  }
+
+  render() {
+    return <DeterministicLinePlot {...this.props}/>
+  }
+}
+
```
abrie, line 117: The goal was a functional solution, but `DeterministicLinePlot` has a lot of state to wrangle. This class intercepts props changes coming from outside the component and debounces the update to 1000msec intervals.

src/components/Main/Results/ResultsCard.tsx
```diff
@@ -191,7 +191,7 @@ function ResultsCardFunction({
         </Row>
         <Row noGutters>
           <Col>
-            <DeterministicLinePlot
+            <DebouncedDeterministicLinePlot
```
abrie, line 194: Use the debounced version.

src/components/Main/Main.tsx
```diff
@@ -130,7 +130,7 @@ function Main() {
     }
   }, [autorunSimulation, debouncedRun, scenarioState, locationSearch, severity])
 
-  const [setScenarioToCustom] = useDebouncedCallback((newParams: AllParams) => {
+  const setScenarioToCustom = (newParams: AllParams) => {
```
abrie, line 133: Remove the useDebounceCallback here because the debounce is now handled by `DebouncedDeterministicLinePlot`.

abrie:
#437

Issue #384, opened 2020-04-03
Improve styling and content of the landing page


ivan-aksamentov:
# 🙋 Feature Request

This ^.
Basically Landing page is empty and not at all like the rest of the app. We need some artistic touch there. Bring some life into it. :)

## 🔦 Context

Importance: will be included into the new release / announcement of the new domain

## 😯 Describe the feature

All optional, non-exhastive (use your imagination!):
 - [ ] SVG should be embedded into the page, right now it loads with a delay on slow connections
 - [ ] Bigger "skip" checkbox, closer to the button. Most users want it checked.
 - [ ] Checkbox should not disappear on the next page load if checked.
 - [ ] Color- and shape-wise should look more like the simulator itself
 - [ ] We might want the navbar on landing page too
 - [ ] Round button looks very differently from our square buttons
 - [ ] Much larger image, updated for the current looks
 - [ ] Maybe loop-roll screencast .mp4 instead of the image  (for fast network && desktop)?
 - [ ] Maybe a youtube video with @rneher explaining the app (does not exist yet)
 - [ ] More meaningful text on the page (from readme?)
 - [ ] Nice-looking links to the Lab page, Github and Richard's Twitter
 - [ ] Maybe add links to FAQ and about?
 - [ ] Maybe team from readme?

## Related

 - https://github.com/neherlab/covid19_scenarios/issues/107
 - https://github.com/neherlab/covid19_scenarios/pull/163

realdy:
Hello, I looked at this project and was really interested in helping out during this time.

How does this look for a prelim image?
![PrelimImage](https://user-images.githubusercontent.com/45107348/78416627-1271ab80-75f8-11ea-80c2-4e6f29b32e72.png)

ivan-aksamentov:
@realdy I was hoping for more (not less) colors and larger image/video - 1/4 of screen or so.
Maybe we don't need the wave. Button should be bigger - this button is the only reason our app exist.

These all are personal preference really. We need an artist here.
Feel free to play with it and revamp the layout completely, if needed.

tryggvigy:
Here is the Figma file I created for the original "design" of the landing page. https://www.figma.com/file/kpT1pGxzqlzCFRYaWyPMSw/COVID-19-Scenarios?node-id=0%3A1

Feel free to fork it and iterate. @realdy I like the colors, top bar and button style you have going on in this iteration.

[...]

PR #414, opened 2020-04-05
384 check style global

kmid5280:
## Related issues and PRs
#384 
 - 

## Description
Moved the style for the new checkbox into the global scss file, so it can be reused elsewhere in the application.

global.scss file, landing page component and scss.

### DIFF
--- a/src/components/LandingPage/LandingPage.scss
+++ b/src/components/LandingPage/LandingPage.scss
@@ -111,64 +111,6 @@
   max-width: 400px;
 }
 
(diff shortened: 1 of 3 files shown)


kmid5280:
Additionally I could put together a smaller checkbox and give it its own class globally. This could be used to replace the default checkbox being used in the results card.

ivan-aksamentov:
@kmid5280 Thanks, but we are happy with bootstrap checkboxes.

PR #426, opened 2020-04-06
[WIP] Feat/read mitigations

rneher:
## Related issues and PRs
preparatory PR for crowdsourced mitigation measures. 

## Description
Use data from countries and regions to prepopulate the scenarios with measures. 

only the data directory. 

## Testing
non yet.


rneher:
@alebrand just sketched this a bit. We'll have to see how we can align this with the JHU initiative.

alebrand:
@rneher I tried mapping the HIT categories to the covid19_scenario ones, could you have a look to the attached file? I made suggestions for extra fields (column C), otherwise many categories end up as "NA".
Note: I extracted the HIT categories from the csv download in the end, as from the codebook, I have the codes, but not the labels reported in the csv file.

We would end up with the following categories in covid19_scenarios:
- Ban on mass   gatherings
- Nurseries/schools/universities closure
- Restaurants/entertainment/shops closure
- Household confinement
- Contact Tracing
- Borders closure/limitation of movements/symptoms screening
- Isolation/quarantine of cases
- Office closures
- Universal mask
- NA

[HIT categories.xlsx](https://github.com/neherlab/covid19_scenarios/files/4440775/HIT.categories.xlsx)

deuteragenie:
> @rneher I tried mapping the HIT categories to the covid19_scenario ones, could you have a look to the attached file? I made suggestions for extra fields (column C), otherwise many categories end up as "NA".
> Note: I extracted the HIT categories from the csv download in the end, as from the codebook, I have the codes, but not the labels reported in the csv file.
> 
> We would end up with the following categories in covid19_scenarios:
> 
> * Ban on mass   gatherings
> * Nurseries/schools/universities closure
> * Restaurants/entertainment/shops closure
> * Household confinement
> * Contact Tracing
> * Borders closure/limitation of movements/symptoms screening
> * Isolation/quarantine of cases
> * Office closures
> * Universal mask
> * NA
> 
> [HIT categories.xlsx](https://github.com/neherlab/covid19_scenarios/files/4440775/HIT.categories.xls

There are many other types of measures or variations thereof.  It would be good to collect an exhaustive list of those, together with a target reduction factor.

Other examples: 

- Outside activities allowed for max. 2 people, with distance > 2 meters
- Nurseries open, but schools closed (so one need to distinguish between both)
- Wearing of mask inside, outside, none, or both
etc. etc. etc.

Ideally, people would package these individual measures into policy decision for each country / dates.

Issue is... some measures are interrelated.

ivan-aksamentov:
Hello!

After discussion with @rneher I put it to milestone 1.2:
https://github.com/neherlab/covid19_scenarios/milestone/3
and mark as important.

What help is needed on the engineering side to make this happen?

cc @nnoll

Vanuan:
Maybe for starters, measure values should be provided in the CSV rather than hardcoded from categories.

Vanuan:
For example, it's reasonable to assume that school closures would have a different effect based on demographics.

ivan-aksamentov:
@Vanuan Thanks for your ideas. Yes, we planned to process JHU data for mitigation interventions, which has detailed information for different countries. After initial prototype with hardcoded data, we could then write a script for that and we would then automatically fetch and update the data every day, similarly to how we do with case counts.

But the entire adventure turned out to be quite tricky and in the end this thing stuck in an undetermined state...

Vanuan:
Here's some content for inspiration: https://www.youtube.com/watch?v=BTLZu-1IMcE

I think for starters we could try gathering 3 sources of data:
1. Manually crafted data of different mitigation measures for different territories (from wikipedia or elsewhere)
2. Apple/Google mobility reports
3. A set of values for manually fitting the prediction curve to actual data

Only after that we can start categorizing different mitigation measures and building a prediction model of their impact.

So, how about getting this file https://github.com/neherlab/covid19_scenarios/pull/426/files#diff-41177582a7c92a97242de0047786712d , adding a column for manual override value and using this piece while throwing out everything else:

https://github.com/neherlab/covid19_scenarios/pull/426/files#diff-9b3e50fca1abc2b9c6ba30b673fd737dR220-R232

PR #433, opened 2020-04-06
Expand the cases through time graph via a new window

gj262:
## Related issues and PRs

Related to #284 

## Description

Adds the ability to detach the cases through time graph into its own window where it can be expanded for a better scale view.

I think this could be merged as is, but rendering is not perfect, the graph will not always re-render in the new window as it is resized. However as parameters are changed and run() is activated it properly updates to new result.

Cases through time graph

## Testing

Pop out, resize, interact, pop back.


ivan-aksamentov:
@gj262 I fixed the deployment error. Please rebase on master.

gj262:
So the resize issue here is similar to what we ran into with popovers when we first tried window.open(). `<ReactResizeDetector />` operates in the context of the parent window and does not notice resize in the child window.

gj262:
Hi @ivan-aksamentov, yeah, this PR is a drift from the original feature request following the drift in the issue. I'd be happy to close this PR and work on the original feature request if that is still an important issue.

> Popup windows is a bad practice. Last time I saw it in 1999. I think any adblocker nowadays, and maybe even firefox by default, will block these from appearing. Plus, we don't want to drive user out of our page.

You may be right that it'll have usability issues in practice (different browsers, blockers). I like the fact that you can tweak the parameters in one window and see the large graph respond in another.  So maybe it is not a feature for everyone but might be a feature for someone. But that someone is only me so far and I'm not an end user of the app so I would not advocate for merging this unless a real user likes it.

ivan-aksamentov:
@gj262 I probably missed the "original" PR. Can you point me to it?

PR #436, opened 2020-04-06
Use generated types within data directory.

nnoll:
## Related issues and PRs

Related to #355 but uses the Python classes from our one source of YAML truth.

## Description
I renamed the generated_types folder because I hated the redudant import statements it caused. All hand-coded classes within Python now directly derived from their corresponding generated/types.py. Furthermore, I added the hard-coded default epidemiological parameters to ~~the YAML~~ an external python file for more ~~central~~ transparent annotation. 

scenarios.py

## Testing

Run the makefile to see that it still works and no output is changed.


data/scripts/scenarios.py
```diff
@@ -101,99 +104,99 @@ def from_ms(time):
...
 # Functions
 
-def marshalJSON(obj, wtr):
+def marshalJSON(obj, wtr=None):
     """ Validate and store data to .json file
     Arguments:
     - obj: a dict of allParams
     """
+    if wtr is None:
+        return json.dumps(self, default=lambda x: x.__dict__, sort_keys=True, indent=4)
```
rneher, line 195: the argument `self` should probably be `obj`.

nnoll: You are correct. It's fixed in the latest commit.

noleti:
Sorry, I don't have bandwidth for a full review today. As I am not really involved in anything but case-count generation, I don't think there will be issues from my side.

PR #437, opened 2020-04-06
Use formik values to render the mitigation table

gj262:
## Related issues and PRs

For #382 

Relates to #432 #412 #403 

~~Note: This is a WIP. It is how Formik could/should be used. But still may not fix the underlying performance issues due to other bugs etc.~~

## Description

Use formik in a less controlled manner.

Mitigation Table

## Testing

Add/Remove from the mitigation table.

### DIFF
--- a/src/components/Main/Main.tsx
+++ b/src/components/Main/Main.tsx
@@ -166,7 +166,7 @@ function Main() {
           onSubmit={handleSubmit}
           validate={setScenarioToCustom}
         >
-          {({ errors, touched, isValid, isSubmitting }) => {
+          {({ values, errors, touched, isValid, isSubmitting }) => {
             const canRun = isValid && severityTableIsValid(severity)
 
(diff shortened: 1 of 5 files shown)


abrie:
Wow this great – using the `values` param from Formik is absolutely the correct approach. This raises my opinion of Formik.

So now there seems to be a path toward decoupling `scenarioState` from the form and perhaps even removing it as a prop dependency...

gj262:
> So now there seems to be a path toward decoupling `scenarioState` from the form and perhaps even removing it as a prop dependency...

This could definitely be done in some areas so that ppl don't try to use it, however scenarioState drives the "Custom" labels on the scenario selector and the age distribution selector, so it needs to remain in place at least for those..

gj262:
@abrie I removed some of the scenarioState per your suggestion (where it was possible to do so).

PR #447, opened 2020-04-06
feat: do not use snake_case for python

brunorzn:
instead, keep original names.

## Related issues and PRs

 -


ivan-aksamentov:
As discussed on Slack, we can return to here sometimes, when we know how to produce the correct Python code with that.

PR #449, opened 2020-04-06
Change ODE integration method

nnoll:
## Related issues and PRs
This is ultimately setting the stage for a revamped PR related to PR #375. A major usability concern with running our algorithm with parameter uncertainty is it introduces noticeable lag due to performance bottlenecks of having to run the algorithm multiple times.

Part of the issue is we currently have to integrate with a small step size as we are using a simple Euler scheme. This PR will change to the more standard 4th order RK integration allowing us to take much bigger step sizes with no loss of numerical accuracy.

## Description

The prerequisites for such a change is to separate the logic that computes time derivatives from the logic that updates the state vector. The output is expect to change a small amount but not qualitatively. The algorithm should run much faster.

model.ts

## Testing

Check that results are the same before and after.


nnoll:
We get a reasonable speed up if we choose half a day steps. With the O(h^5) error I think this is reasonable. Test snapshots will need to be updated as they are slightly different. I think we are starting to hit diminishing returns with performance optimization in the browser...

rneher:
I like your implementation. I think if we wanted to optimize this further, we would need to replace the loops over age keys with maps over arrays. Not sure how much that would save, but could be a factor 2-3.

rneher:
Did you quantify how much it sped up things? we have a 10x longer time step, but probably 5x more complex calculations. so not really obvious.

nnoll:
It was a 2x speedup on my machine (taking the average of like 5 runs before and after). I like your map idea as we can then remove the reliance on the keys within the algorithm

nnoll:
OK moved away from summing over a dictionary to more natural arrays in the hot loop and got a big speed up. I think this looks promising to merge!

rneher:
Yes, this looks very good to me. The only thing that I am a little wary of is the extent to which we rely on dictionary order. does 
`severity.forEach((row, i)`
and 
`Object.keys(ages)` 
necessarily have the same order? might be good to guard against nasty surprises.

nnoll:
OK. I've explicitly included a step that sorts keys in both the severity table and age-distribution lexicographically. The only surprise that could happen now is if we are handed keys that don't have the property that sorting lexicographically = sorting numerically. But this is outside the scope of this PR I think...

Issue #442, opened 2020-04-06
Spinner is not centered


dkozar:
## 😯 Current Behavior

![spinner](https://user-images.githubusercontent.com/1078403/78603779-9d4ce300-7850-11ea-897a-a6f14126a815.png)

## 🤔 Expected Behavior

![spinner2](https://user-images.githubusercontent.com/1078403/78603797-a2aa2d80-7850-11ea-9609-f1c377662308.png)

gj262:
Improved by @kmid5280 🎉.
Still not quite centered though.

PR #453, opened 2020-04-07
Center spinner

kmid5280:
## Related issues and PRs
#442 
 - 

## Description
Move spinner farther down page. Set margin-top property in spinner container to 25%. Recommend testing with different browsers and sizes to make sure desired effect is achieved.

Loading.scss

## Testing
Open main simulator page. Refresh page to view spinner.

### DIFF
--- a/src/pages/Loading.scss
+++ b/src/pages/Loading.scss
@@ -1,5 +1,5 @@
 .loading-page {
-  margin-top: 1rem;
+  margin-top: 25%;
 
   > div.spinner-border {
     width: 6rem;


gj262:
thanks @kmid5280 a definite improvement!

kmid5280:
@gj262 In this case I just moved it a little further down the page. Another possibility may be to make it dynamic so it centers depending on the browser height.

dkozar:
I think the best solution would be to use viewport width and height.

Pseudo code:

```
@spinnerSize = 100px // just assuming

.spinner {
    position: fixed
    top: calc((100vw - @spinnerSize)/2)
    left: calc((100vh - @spinnerSize)/2)
}
```

Viewport units: https://css-tricks.com/fun-viewport-units/

Issue #233, opened 2020-03-26
Improve readability of the results graph tooltip


ivan-aksamentov:
# 🙋 Feature Request

### Presence
Currently lines in the tooltip appear and disappear depending on a value of a particular point on x axis. Making focusing on a particular line difficult. In particular, a line "ICU overflow" disappears when zero. Same with all "Confirmed" lines.

### Color
Additionally, some of the entries are colored such that they don't have a sufficient contrast on the white background. This somehow happens with the most important entries, such as "Deaths" and "Confirmed *".

## 🔦 Context

Some work is done in https://github.com/neherlab/covid19_scenarios/pull/157
After it merged, we can take a look at further improvements.

## 😯 Describe the feature

 - [x] All lines should be present at all times when hovering different parts of the plot.
 - [x] "Confirmed *" lines should say "-" as a value when a value is not present
 - [ ] (optional) "Confirmed *" lines  should be shown only if there is at least one value present in any of these lines. But they should be either all shown or all hidden. This is for regions without confirmed data.
 - [ ] Other lines should show "0" when a value is not present. For example "ICU overflow".
 - [x] Revamp color scheme to be more accessible (improve contrast ratio fg-to-bg)
 - [ ] (desirable) Make color scheme accessible for color blindness (use online generators and verify in Firefox https://developer.mozilla.org/en-US/docs/Tools/Accessibility_inspector/Simulation)

## Related

 -  https://github.com/neherlab/covid19_scenarios/pull/157 (implements better positioning on mobile and arranging lines in 2 columns)

kmid5280:
@ivan-aksamentov Regarding making the color scheme accessible, I found a way to make the lines dotted. We could potentially take similarly-colored lines and make them dotted or solid, or find other ways to differentiate them. Would that be a good start?

![dotted lines](https://user-images.githubusercontent.com/35879775/77815254-d9c55580-707e-11ea-9770-90e1a470a2f5.png)

ivan-aksamentov:
@kmid5280 Hm, I am not sure.  Dotted lines change semantics, so that should be coordinated with our science team.
I summon @rneher  and @nnoll ! :sparkles:

kmid5280:
@ivan-aksamentov Any further thoughts on this idea so far?

gj262:
I will take a look at the first 4 tasks of this unless someone else is working on it? 
It should also help alleviate #234

ivan-aksamentov:
Mostly resolved by #495

PR #471, opened 2020-04-07
Feat/print pdf

brunorzn:
## Related issues and PRs

 Contributes to https://github.com/neherlab/covid19_scenarios/issues/110

## Description

This pull request uses mostly CSS rules to prevent things to be printed out - or exported as a PDF using the browser dialog.

It has nothing to do with what has been already done in the open PRs. It doesn't load any extra library or anything, it just uses CSS.

The only thing that is impacted outside of specific-to-print is the export dialog : I've added a "print" button there.

## Testing

Run a simulation > Export > Print.

I've tested it on chrome and it works very well.
I have tested it on Firefox (Mac OS) and only the first page would be exported. Maybe a problem with some META Tags ? I've read that somewhere, but I can't find it anymore. I don't know, I didn't have time to fix this, but let's hope it will be easy.

Example (printed from chrome) : 
[COVID-19 Scenarios.pdf](https://github.com/neherlab/covid19_scenarios/files/4447469/COVID-19.Scenarios.pdf)


brunorzn:
Update :

it doesn't work with safari either (which displays two columns)
and it's a bit ugly on Windows

I'm probably just too bad with CSS.

I still think it's a good approach, and I hope things can be fixed quickly, at least on chrome for every platform. 

I'm letting this PR open, so you can try things out on the preview.

gj262:
In case anyone didn't know this (me for instance):
https://css-tricks.com/can-you-view-print-stylesheets-applied-directly-in-the-browser/

ivan-aksamentov:
@brunorzn Good job!

The layout is still quite janky, but this can be improved, I think. Firefox/Safari would be much more challenging.

However, my primary concern is whether the contents of this PDF is of any use.

Our main results are the numbers such as deaths, number if patents in ICU, ICU overflow in the given time. The essence of the app is in these numbers. That's what we need to communicate.
However, we don't explicitly show them in the app. User have to hover the plot to see these numbers in a wonky and jumpy tooltip with fields constantly blinking on and off ( #233 ). In PDF it's even worse - no tooltips - numbers cannot be seen. In the end, there are 5 pages with almost no useful results.

We should find a way to communicate our results more clearly. Maybe we need to figure some sort of a summary table? Maybe print-only for now?

The challenge is to find an appropriate visual representation for the time-series. Should we chose a fixed step? Should user decide?

cc @rneher @nnoll @noleti @gj262 @abrie

rneher:
yes, it works reasonably well on chrome, but on firefox only the preamble is present.

noleti:
I tested on Linux/Firefox 74.0.1, and I also just get the disclaimer page 
[example(https://github.com/neherlab/covid19_scenarios/files/4448990/covid.pdf)

brunorzn:
I'm working on fixing layouts for both Firefox and Chrome (not mac).

Once this is done, we can implement printed-only figures.

brunorzn:
Update:

exports now works for both Chrome / Firefox but the process is a bit different.
in the "export" dialog, you can generate a preview, which will clean the dom up before printing.

that way, it's much easier to have control over what's being printed or not (no more columns, floats, etc).

I've also added labels to some diagrams.

[CHROME.pdf](https://github.com/neherlab/covid19_scenarios/files/4451167/CHROME.pdf)
[FIREFOX.pdf](https://github.com/neherlab/covid19_scenarios/files/4451169/FIREFOX.pdf)

brunorzn:
updated, with table.

[withTable.pdf](https://github.com/neherlab/covid19_scenarios/files/4451488/withTable.pdf)

PR #476, opened 2020-04-08
combine the age chart and the peak value chart. Age dis missing from …

rneher:
…tooltip

## Description
We currently waste a lot of space on displaying the age distribution which is an input to the model, not an output. It is useful as a comparison though. This PR includes the age distribution into the bar chart with peak severity, fatalitie etc

AgeBarCart.ts

## Testing
Check whether this is more informative.


nnoll:
This is a large improvement.

rneher:
do you guys prefer legends above or below the plot? I always felt stuff is too cluttered with them above.

nnoll:
Below

Issue #461, opened 2020-04-07
Address bar should not contain the sharing parameters


ivan-aksamentov:
# 🙋 Feature Request

I propose to:

 - [ ] only generate sharing URL on explicit user action (clicking "copy" button)

 - [ ] disable synchronization between app state and current URL parameters in the address bar

 - [ ] remove incoming URL parameters from URL on navigation to shared URL

 - [ ] introduce explicit sharing action - through email, twitter and other social media eventually #462 

Motivation:

 - reduce amount of bookmarks and links hanging in the internet that contain parameters unnecessarily (with no intent on persisting or sharing them)
 - facilitate onboarding and reduce the confusion for new users. Their bookmark should be to the `/`, not some random parameters (with bogus results) they played in the first visit
 - reduce chance of failures in case where we don't handle old parameters correctly at some point in the future (and it WILL happen)

## Related

 -
 -

ivan-aksamentov:
Resolved in #478 #481

PR #478, opened 2020-04-08
#461/Remove autoupdating scenarioURL.

abrie:
## Related issues and PRs

Contribtutes to #461, but does not implement 'sharing' feature. 

## Description

This remove auto-generation of the URL parameters.

WIP:

- ~~App needs to parse URL if present, and apply to scenario params.~~
- ~~Run in new tab needs to generate URL before opening window.~~
- ~~Run in new tab should be disabled if simulation cannot be run~~

Browser URL should be clean, no longer polluted by URL params.
"Run in tab" internal logic has changed, but should look unchanged to the user.

## Testing

1. Open a fresh browser session
 - Confirm there are no url params, i.e. URL should simply be `http://for.example.com/`
2. Change some scenario parameters.
 - Confirm there are no url params.
3. Run the scenario.
 - Confirm there are no url params.
4. "Run in new tab"
 - A new tab should open, and the simulation should be run.
 - The new tab should have no url params (they may flash momentarily, then be cleared away.)

### DIFF
--- a/src/components/Main/Results/ResultsCard.tsx
+++ b/src/components/Main/Results/ResultsCard.tsx
@@ -31,6 +31,7 @@ interface ResultsCardProps {
   severity: SeverityTableRow[] // TODO: pass severity throughout the algorithm and as a part of `AlgorithmResult` instead?
   result?: AlgorithmResult
   caseCounts?: EmpiricalData
+  getScenarioUrl: () => string
 }
 
 function ResultsCardFunction({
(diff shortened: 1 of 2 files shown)


src/components/Main/Results/ResultsCard.tsx
```diff
@@ -131,7 +131,7 @@ function ResultsCardFunction({
               <LinkButton
                 className="new-tab-button"
                 color="secondary"
-                disabled={!scenarioUrl}
+                disabled={!canRun}
```
ivan-aksamentov, line 134: Could it be that we canRun but the scenarioURL is not ready?

abrie: Good question. The app will immediately try to run the scenario provided in a URL link, so it makes sense to ensure that it is indeed runnable before enabling the run in new tab.

PR #479, opened 2020-04-08
Model uncertainty in input parameters

nnoll:
## Related issues and PRs

An updated version of PR #375 but now living in the validated world of YAML. Discussion #269.

## Description

Run has been changed to iterate over this array and computes an average of both the linear and quadratic terms over all stochastic realizations to allow for easy return of the mean and variance of outcomes.

## Testing
Tests will need to be written for the affected areas. I'll start this as the PR solidifies.

## Requested help

The new form found at src/components/Form/FormRangeSlider is my sketch at the final widget. I view it as a prototype that works on my screen (there are hard-coded values here). My burgeoning React skills are at their limit to make this a robust component that we could use elsewhere potentially, e.g. within the mitigation values. I would like a pro to look at this.


nnoll:
A review of this branch would be welcome (not for merging but just so we can see what does/doesn't work here)

rneher:
I played a bit with different ways of averaging. Given the exponential nature of the process, an arithmetic average seems no optimal. Having these deeply nested structures makes these averages a little cumbersome -- have a look

nnoll:
I agree your point. Computing the rolling averages was a bit of unneccesary premature optimization of memory usage. I think since we store the trajectories it would be worthwhile to also use the IQR instead of variances as this would be even more sensitive to exponentials...

nnoll:
I also agree about the clunkiness about summing over a bunch of these structs. This is where operator overloading would make code look cleaner. I'll tinker on your changes because I think we can reduce a bit here. I have defined sumTimePoints(x, y, func) that computes z = x + func(y). This solves the summation procedure for the geometric mean computation but not the exponential rescale. I think if we seperate the above function into

(1) sum(x, y) -> z
(2) scale(x, func) -> y

We could make the code more transparent. I'll do this now.

nnoll:
OK, I played with this some more today. The main changes were that I made it so error bars are no longer symmetric and thus asymmetric error bars are reported in our interface. Additionally, there was a good deal of refactoring to make the operations more clear and easy to modify. Lastly, UserResult has an optional field where we could attach percentiles of a more detailed run (the function of which is already written, we just have to call it somewhere). I think we should leave the implementation of that for another PR because this one is getting quite large...

rneher:
Agreed with leaving the percentiles for a little later. If we wanted to merge this sooner rather than later, we might want to revert the form to a simple number field while we work more on uncertainty in the background. I think the uncertainty in the measures is quite critical to assess future scenarios. 
(we should also rethink the `log(x) || -20` -- it says `+20` somewhere and this should at least by a variable with a name like `logPseudoCount`. We could also add smth to the argument of the log)

nnoll:
I like adding the term to the argument of the log as this makes more sense to my brain than the or notation of javascript. I think we only need the regularization at the site of the log call so I've removed it elsewhere.

Merging this while removing the slider might make sense, however a lot of the new code relies on the new type of r0 being a tuple...

rneher:
@nnoll now that you have a separate plot for the mitigation, should we add a curve for the resulting r0?

ivan-aksamentov:
Continued in #537

PR #454, opened 2020-04-07
#443/Replace default numeric input spinners with suitable alternative

abrie:
## Related issues and PRs

Fixes #443, and expands scope to _all_ numeric interval spinners.

## Description

Replaces interval value spinners with a custom component that supports fine-grained values while maintaining step size.

~~Currently using [react-numeric-input](https://github.com/vlad-ignatov/react-numeric-input).~~

Switched to a custom component after several unsuccessful trials of third party components. https://github.com/neherlab/covid19_scenarios/issues/443#issuecomment-610063175

## Impacted Areas in the application

All ~~Mitigation interval's~~ numeric input spinners.

## Testing

Check all spinner-controlled fields on multiple browsers...

PR #480, opened 2020-04-08
#475/Validate before state update

### LINKED ISSUE #475: Form validation no longer works
# 🐛 Bug Report

Form validation no longer works. It has to be fixed ASAP.
There are no validation errors shown. Invalid values (for example negative) are silently accepted and passed to the algorithm.

## How to reproduce

Steps to reproduce the issue:

 2.
 3.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Browser          |
| Operating System |
|                  |

## Related

 -

gj262:
## Related issues and PRs

Fixes #475 

## Description

The App uses Formik's validate method to trigger autorun. However it does not validate the date. And so the state used to run() is possibly invalid. 

This change uses the yup schema to validate the parameters and only updates state if they pass.

## Notes:

1. Mitigation table does not display validation errors. Perhaps that is fixed in #454 
2. There is double validation input to Formik. However removing `validationSchema=` disables validation. Possibly this is due to the debounced nature of `validate=`. 

Form validation
Auto run

## Testing

```
with URL ingestion/update disabled

clear localStorage
load /
Set latency = -1
Focus out of input
=> Should be non-negative error displayed
Run is disabled
PASS

Enable auto run
run() is not triggered
PASS

Set latency = 4
run() is triggered with latency = 4
PASS
```

### DIFF
--- a/src/components/Main/Main.tsx
+++ b/src/components/Main/Main.tsx
@@ -3,7 +3,7 @@ import { useDebouncedCallback } from 'use-debounce'
 
 import _ from 'lodash'
 
-import { Form, Formik, FormikHelpers, FormikErrors } from 'formik'
+import { Form, Formik, FormikHelpers, FormikErrors, FormikValues } from 'formik'
 
 import { Col, Row } from 'reactstrap'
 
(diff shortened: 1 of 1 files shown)


gj262:
> @gj262 Validation errors seem to be appearing correctly.
> However, autorun still happily runs with invalid values.

I did not see this. I disabled URL update while I tested this change in anticipation of #478. Now that that is merged can you retest?

ivan-aksamentov:
@gj262  Let's merge and monkey-test if it reappears

Issue #474, opened 2020-04-08
Fix truncation of locale dropdown


ivan-aksamentov:
# 🐛 Bug Report

Fix truncation of locale dropdown:

![Screenshot from 2020-04-08 14-14-15](https://user-images.githubusercontent.com/9403403/78783096-5c191800-79a3-11ea-88d6-e8f8f0a20c08.png)

***IMPORTANT*** Navbar items should be kept scrollable on mobile.

## How to reproduce

Steps to reproduce the issue:

 1. Open the application in a browser
 2.
 3.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Browser          |
| Operating System |
|                  |

## Related

 -

guandiyiyi:
sent a PR for the fix

ivan-aksamentov:
Resolved in #484

PR #484, opened 2020-04-09
fix issue#474: the language switcher is cut off by the main content

guandiyiyi:
## Related issues and PRs
issue #474 
 - 

## Description
The language dropdown is cut off by the main content, because it's attached to navigation container, add attribute positionFixed to fix the menu position.

NavigationBar.tsx
LanguageSwitcher.tsx

## Testing
open the application, click the language switcher, noticed the dropdown is shown as expected.
image


rneher:
yes, we disabled this for the time being since there are so many strings that are untranslated that the different languages are not very useful.

guandiyiyi:
Ah, ok, I saw this issue tagged as Important such that I picked it up, also I have another PR fixing translation of date picker. You can take a look as well.

whiver:
So is it a real bug or is it expected? Because this I'm working on improving the translation process in #494, so this would be a good opportunity to improve the translations, if we re-enabled them 😁

ivan-aksamentov:
@whiver This is a chicken and the egg kind of problem. We cannot enable i18n before there is enough translation. And we cannot ave these translations before people come and translate.

Also we tend to change strings all the time. Automation should help, hopefully.

PR #485, opened 2020-04-09
Fix/tslint

brunorzn:
## Description

Since the release is done, this PR resolves most of TS/TSlint issues.
The only ones remaining are _Refactor this function to reduce its Cognitive Complexity_ in

* Main.tsx
* DeterministicLinePlot.tsx
* SeverityTable

This PR mostly solves a couple of issues regarding types :
* SeverityTableRow is now in ScenarioTypes (it is used in many files, and it also solves a cyclic dependency issue)
* OneCountryAgeDistribution was still used in several places, but it doesn’t exist anymore !

## Testing

yarn tslint


src/components/Main/Main.tsx
```diff
@@ -116,19 +119,13 @@ function Main() {
...
       updateBrowserURL('/')
     }
-  }, [])
-
-  const [debouncedRun] = useDebouncedCallback(
-    (params: AllParams, scenarioState: State, severity: SeverityTableRow[]) =>
-      runSimulation(params, scenarioState, severity, setResult, setEmpiricalCases),
-    500,
-  )
+  }, [autorunSimulation, debouncedRun, scenarioState, severity]) // eslint-disable-line react-hooks/exhaustive-deps
```
gj262, line 122: The current version `[]` gives run once (componentDidMount) semantics which is what I think we want to limit the LocalStorage and URL futzing.

gj262:
This is great! It will make working in the codebase that much easier and less error prone. (Though on a personal note I will miss `OneCountryAgeDistribution` ;)).

src/components/Main/Main.tsx
brunorzn: Should I set it to [] then ?

gj262:
Yeah, it should be [].

brunorzn:
Okay, I've updated the useEffect dependencies @gj262

gj262:
@ivan-aksamentov Can I merge this? It will be a good addition. And because it touches so many files it would be good to get it merged sooner rather than later.
CC @rneher

Issue #440, opened 2020-04-06
Translate date input boxes and date picker widget


ivan-aksamentov:
# 🙋 Feature Request

Translation is currently incomplete, and, in particular the Airbnb's DateRangePicker is not connected to the i18n part of the app.

## 😯 Describe the feature

Date picker and date input box text should be translated according to currently selected locale.

## Related

 -
 -

dkozar:
![image](https://user-images.githubusercontent.com/1078403/78671579-65d54980-78d7-11ea-9735-e2475b744f58.png)

guandiyiyi:
sent a PR for fix, root cause:
the locale of moment is not set when language change

PR #486, opened 2020-04-09
fix Issue440: translate moment strings in date picker and date dropdown when switch language

guandiyiyi:
## Related issues and PRs
issue #440 
 - 

## Description
The locale of moment is not set when switch language, which cause the date picker and date dropdown is not translated according to local.
Set locale in moment when switch locale in language dropdown.

languageSwitcher
createMitigationInterval
DateRangePicker
TableResult
ScenarioCardEpidemiological

## Testing
click date dropdown in scenario card -- all the strings translated
click date picker -- all the strings translated
switch language -- all the strings translated to the selected locale
image
image

### DIFF
--- a/src/algorithms/utils/createMitigationInterval.ts
+++ b/src/algorithms/utils/createMitigationInterval.ts
@@ -1,4 +1,4 @@
-import moment from 'moment'
+import moment from 'moment/moment'
 import { v4 as uuidv4 } from 'uuid'
 import createColor from 'create-color'
 
--- a/src/components/Form/DateRangePicker.tsx
+++ b/src/components/Form/DateRangePicker.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react'
 
-import moment from 'moment'
+import moment from 'moment/moment'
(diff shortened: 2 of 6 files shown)


ivan-aksamentov:
@guandiyiyi Overall looks like the right direction, however:

 - importing es5 version of moment increases bundle size dramatically (see debugbear checks below or run `yarn analyze` before and after change to see the diff)
 - the "more correct" way is to load only locales that we use (can be done in webpack, the context replacement plugin is already there)
 - to load locales asynchronously when they are needed would be fantastic
 - Polish months are showing in some strange `(one|two)` format. I think that thre  obtained values should be formatted with `.format()` additionally

Could you please rebase and look into these?

P.S. It is better to not mix two issues into the same PR

guandiyiyi:
Thanks, I will take a look these issues the weekend. I have the fix of issue 474 in this PR as well cuz I need add the language switcher back for testing, and forgot to exclude it out. I have another PR actually sent for issue 474, could you take a look? Once language switcher is added back, I will merge master and remove the unnecessary part. Thanks.

Issue #397, opened 2020-04-04
Replacing the scenario dropdown with "Load scenario" dialog


dkozar:
# 🙋 Feature Request

Let's replace this dropdown with a "Load scenario" button:

![image](https://user-images.githubusercontent.com/1078403/78424166-71203f00-7663-11ea-8571-1b1a964db4fb.png)

## 🔦 Context

There is a general confusion about what this dropdown means and how does it relate to parameters, once the parameters are changed.

Also, there's no clear vision on **should we serialize this parameter** to query string in the URL.

My conclusion is that this actually isn't one of the input parameters. It's just a way of loading an existing scenario. So, let's remove the dropdown because this is not a parameter.

If we need to have a scenario name on top of the page (for sharing purposes etc.), then we can have a text field with scenario name and text area for description.

## 💁 Possible Solution

Button:

![image](https://user-images.githubusercontent.com/1078403/78424199-a2007400-7663-11ea-9b21-050570edc221.png)

Dialog:

![load-scenario-2](https://user-images.githubusercontent.com/1078403/78424205-a9c01880-7663-11ea-9f2f-83918500d2bf.png)

This would make a "scenario picker" a rich control, providing much more context (author, description etc.), which is great as the number of scenarios grows.

There would also be a "Search" field on top, and possibly some user filters, pagination etc. There could be a filter for seeing only "My scenarios", or the ones belonging to "My org".

rneher:
I agree, this is more intuitive. these scenarios are really just presets that can be tweaked. I like the proposal.

dkozar:
Let's engage the UX and implement it. It should be pretty simple.

nnoll:
I just saw this after the opened PR. I really like this direction, particularly the potential flexibility with loading presets or even as a way to distribute narratives. The scenario name is ultimately not a parameter but rather a way for people to get to the simulation they care about. This could honestly go on the now deprecated landing page...

Also regarding serializing the scenario name, we should not serialize this. It is ultimately not a parameter as you stated.

dkozar:
Yeah, the dropdown is really confusing, even to us contributors ("is it one of the params or not?").

It's the first usability issue the user encounters in the app; let's fix it.

ivan-aksamentov:
Resolved in #490

PR #490, opened 2020-04-09
feat: Preset loader

dkozar:
## Related issues and PRs

 - https://github.com/neherlab/covid19_scenarios/issues/397

## Description

471882FA-0139-4020-88DE-AABEF3F0CDE0


nnoll:
This is pretty slick. My only comment on this is it would be nice to still display which scenario you are looking at somewhere.

dkozar:
> This is pretty slick. My only comment on this is it would be nice to still display which scenario you are looking at somewhere.

Please do suggest (specify) what data needs to be displayed in the table, I’ll add it to the table. Or whatever else is needed and I don’t understand.

noleti:
> > This is pretty slick. My only comment on this is it would be nice to still display which scenario you are looking at somewhere.
> 
> Please do suggest (specify) what data needs to be displayed in the table, I’ll add it to the table. Or whatever else is needed and I don’t understand.

I think the point is the following: After I select a scenario (e.g. Germany), the name of the scenario is not displayed in the app any more. I see that the age distribution is from Germany, and the Cases displayed are from Germany, but not that the scenario name was Germany. 

Lets assume we will have multiple scenarios per country in the future, one with high and one with low R0. In the current state, it will not be possible to tell which of the two was loaded after the preset loader is closed.

dkozar:
I see. So there should probably be a big title on the top, which one can change. For instance, a little edit button should turn it into an editable field. Just like the title of this GitHub issue could be changed.

The title would also be persisted in the URL.

Also, in the future “presentation mode”, the title should probably render above graphs.

What do you think?

dkozar:
Unite tests:

tests

dkozar:
button


dialog

dkozar:
PR is now ready for review.

As for the scenario title (discussed above) - this will be added in a separate PR, after we discuss where.

The field "current" is still being de/serialized and this PR doesn't change that.

dkozar:
Making labels clickable:

image

dkozar:
@nnoll @noleti what do you think?

https://covid19-scenarios-nd9y08rft.now.sh/

image

dkozar:
I would then make the title editable, just as this page's title:

image

image

noleti:
Looks great to me. I'll leave the merging to the others, though.

ivan-aksamentov:
I am strongly against.

This is no different from dropdown functionality. 
I, personally, don't like the UX - it's slower, more annoying and less accessible (context switching for the popup, mouse clicks need to be targeted on entries, in keyboard mode - more interactions and tabbing) 

If scenario name needs to made more clear then the additional large heading like on screenshot should do. Dropdown can be made bigger in needed. But honestly, none of this is of a concern.

Ultimately, this is just another unnecessary blob of code I need to review, merge and maintain for no reason. We never asked for this. We have important features hanging not implemented for weeks now.

I will rely on Richards and Nick's opinion there.

rneher:
I do think this has value in that it separates presets from parameters from a UI perspective. My suggestions would be to change the button to "Load different scenario" and maybe include the name of the scenario into the card title. Currently, the title uses up a lot of space.

nnoll:
@dkozar @noleti interpreted me correctly. I just wanted an indication of the option I selected at the beginning. This looks great, although I do agree it takes up a bit of real estate right now. 

I am also all for visually separating the scenario name, which we only use as a dispatch into a table of presets, from the actual input parameters to the model. I don't think this ever needs to get serialized inside the URL as all of the parameters that are serialized supercede this information. We could just default to "Custom" or "Linked" for URL sharing. Depends on how much weight this adds.

I'm not qualified to speak on amount of code added from a maintainence perspective, but I can say a "context switch" for exploring different regions is a net benefit from a UX perspective. Changing locations we run the simulation for should "feel different" relative to switching the assumed R0 value for example.

dkozar:
@rneher I changed th ebutton text to "Change scenario".

As for the title, let's consult the UX for the best option and than move it (it's trivial).

@nnoll as for the serialized scenario name - it turns out it's handy, although it's not part of params. It's "meta" param, a custom text. It would be great if the user could change their custom scenario title to "Scenario XYZ" and send this link around. The opener would see the same scenario name. What do you think?

nnoll:
@dkozar I like the suggestion as it allows one to catalog different simulations. This name could even be added as a prefix onto the output file names.

dkozar:
@nnoll also, the filtered preset list could be made a component and reused. Someone mentioned that the landing page might show the list of all the scenarios (?) So the user could find the one they are interested in (full page, as in Google search) and load it.

nnoll:
Hey @dkozar, we discussed this and think this would be great to include within Milestone 1.2. We were thinking about the following changes/additions:
 - [ ] Customizable scenario name. This would be useful for file export.
 - [ ] Adding an import option that could eat an exported JSON file.

What do you think?

dkozar:
Sure, it’s easy to do.

dkozar:
> Hey @dkozar, we discussed this and think this would be great to include within Milestone 1.2. We were thinking about the following changes/additions:
> 
> * [ ]  Customizable scenario name. This would be useful for file export.
> * [ ]  Adding an import option that could eat an exported JSON file.
> 
> What do you think?

I created the issue here: https://github.com/neherlab/covid19_scenarios/issues/541

PR #494, opened 2020-04-09
Automatically extracts the localized strings from the code

whiver:
## Related issues and PRs
- Contributes to #56 

## Description
Add a new script that extracts all the strings from the code and populate the english localization file with them (locales/en/translation.json). This file can be then used as a reference to translate in other languages.

Hopefully this will help us have more accurate and up to date translations.

The script is not run at build time for now as I don't exactly know when would be the best time to run it. You have to run it manually by running `yarn i18n`.

The process of updating the translations to another file can be then either done using a translation platform (best solution, which would ease the collaboration on that) or manually (we could update the script to take the target language as a parameter if necessary).

We can discuss which translation platform to integrate if you want. I just tested GitLocalize and it's very easy to setup, might be a good way to start without too much hassle ([you can see what it looks like on my fork](https://gitlocalize.com/repo/4297)).

Translations.

## Testing
Run `yarn i18n`.


noleti:
This looks amazing imho. I went through the german translation and was able to quickly bring it from 64 to 100%. I didn't review the code changes, but the result is very contributor-friendly.

rneher:
Agree, this looks super useful!

ivan-aksamentov:
@all-contributors please add whiver for code and translations

PR #495, opened 2020-04-10
Tooltip consistency

gj262:
## Related issues and PRs

Contributes to #233 

## Description

This change deals with the _presence_ items of #233. Namely:

- All lines are present at all times when hovering different parts of the plot.
- "* (data)" lines from observations say "-" as a value when a value is not present.
- Lines are shown only if there is at least one value present.
- Other lines show "0" when a value is not present.

Graph tooltips.

## Testing

View the tooltip for the `DeterministicLinePlot` graph for various scenarios.
With "humanized" numbers on/off.
In narrow and wide formats.
The data displayed is consistent for the scenario.
Check for regressions on AgeChart.


rneher:
This looks like a great improvement to me. Both in terms of consistency/interpretability and code structure. We need to make sure that the changes are compatible with

https://github.com/neherlab/covid19_scenarios/commit/8a5d5e61b745d8a21d072126ef5cb2073021327c

which we commited last night to fix an issue with the plot when you change tMin (prev aggregation of data by day failed when dates had different times during the day...)

gj262:
This change will coexist nicely with that.
-Gavin

PR #497, opened 2020-04-10
Updated data from Dominican Republic

medinaman:
Added population, updated cases and deaths and added data from priority province (Duarte).

## Related issues and PRs

 -


medinaman:
Looking forward to have more data from Dominican Republic uploaded to this repository to run your tool.
Thank you.

noleti:
@medinaman I am closing this now, as we now have full populationData, hospital beds, and estimated ICU beds for Dominican Republic from other sources (CIA facts book). So the app now fully supports Dominican Republic without merging the PR. 
Extra entries for DOM-Provincia Duarte would make sense only if someone manually maintains the data .tsv, which does not appear likely at the moment. Alternatively, we would need a parser to get the daily updates from an API. If you have such a parser, or API, feel free to comment here and we can re-open the PR.

PR #505, opened 2020-04-11
Run simulation on a webworker

abrie:
## Related issues and PRs

Contributes to #581

## Description

Runs the simulation algorithm on a webworker.

UI responsiveness improved.
The "Run" button changes to "Running..." while the algorithm runs.

## Testing

~Currently does not run on the staging environment. Likely a production configuration issue(?)~

1. Press 'Run'
  - Observe the button text changes to "Running..."
  - Observe the button text changes back to "Run" when simulation completed.

2. Toggle auto-run
  - Observe the button text changes to "Running..."
  - Observe the button text changes to "Refresh" when simulation completed.

3. Change a scenario parameter
  - Observe the button text changes to "Running..." after ~1 second.
   - Observe the button text changes to "Refresh" when simulation completed.


abrie:
Fails to run on staging environment:

image

ivan-aksamentov:
We were considering webworkers from the beginning, but because the algorithm is sub-50 ms we decided to not bother. I though message passing will be the bottleneck. But now that I see it running, it actually feels much snappier for some reason.

I am particularly worried about our scientists messing with message passing code. Things may derail very quickly. I'd prefer maybe less efficient, but very simply looking interface between the threads, without too much scopes, to avoid silly things.

Would you like to additionally play with
https://github.com/GoogleChromeLabs/comlink
in order to make message passing less hairy?

Also, strangely all pages gained +6.77 KBytes.

src/workers/algorithm.js
```diff
@@ -0,0 +1,16 @@
...
+  const { paramsFlat, severity, ageDistribution, containment } = event.data
+
+  run(paramsFlat, severity, ageDistribution, containment).then((result) => {
+    // Delete function properties, because they cannot be cloned by the structured clone algorithm:
+    // https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
+    // TODO: Revisit this hack.
+    delete result.params.rate.infection
+
+    postMessage(result)
+  })
```
ivan-aksamentov, line 15: This is super hairy :)

We need to find a way to deal with this without deletion. Also, what happens on another end? Is it being restored?

Additionally, could we please use `async/await` if possible?

abrie: Yeah very hairy. The function is not added back because it's not used anywhere in the UI. Poor assumption; but it was a quick hack to a proof of concept. I'm happy to look for a better way to do all this.

ivan-aksamentov: @abrie I wonder if `comlink` can help to marshal/unmarshal the function,

On the other hand, having a function there is weird too. Maybe it has to be rethought.

abrie: A quick look at the comlink documentation mentions structured cloning, which excludes functions: https://github.com/GoogleChromeLabs/comlink/blob/master/structured-clone-table.md. Maybe there is way around that by using proxies, but maybe the function doesn't need to be there in the first place. Summon the scientists for more insight?

nnoll:
Moving forward after 1.2 this will become very relevant as we will start to simulate many scenarios in batch and report the results. So we should consider this seriously

src/workers/algorithm.js
ivan-aksamentov: @abrie Great news!
Nick has removed this function (more precisely the entire params blob) from the results, so this should resolve itself after https://github.com/neherlab/covid19_scenarios/pull/537 is merged. Which should happen right after 1.1 release later today.

abrie: That is great news – now the structure is much better suited for the structured clone algorithm.

Currently there are some problems between the `worker-plugin` and `eslint-plugin' components. As far as I can tell the linter is being run twice, with the second pass throwing errors because it's applied to transpiled code. The resulting errors are breaking the build.

ivan-aksamentov: @abrie eslint plugin will be disabled in inext couple of days and this will run instead:

https://github.com/neherlab/covid19_scenarios/blob/c75018588db67204f072a8f17425c9702332df1b/config/webpack/lib/webpackTsChecker.ts#L68

But first I need to finalize some tricky parts of type checking. I hope that will play better with `worker-plugin`.

abrie: @ivan-aksamentov Just a ping to note that it now looks like `react-refresh-webpack-plugin` is incompatible with `worker-plugin` and probably most alternatives (`worker-loader`, `workerize-loader`). The core issue is that the worker context has no access to the window context, and this is causing problems via HMR. The build works, but the app does not.

ivan-aksamentov: @abrie  This is probably because webworkers have no access to `window` and hot reloading relies on it. Could you please try to set `output.globalObject: 'this'` in webpack config and see if both webworkers and hot reloading in main thread work?

abrie: @ivan-aksamentov I've tried a few variations on globalObject: `self`, `this`, and `false`. The webworker continues to fail:

image

Looks like it may be known issue with `react-refresh-webpack-plugin`:

https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/24

Which was seemingly corrected by https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/29.

ivan-aksamentov: @abrie I have updated `@pmmmwh/react-refresh-webpack-plugin` on master just now

abrie: @ivan-aksamentov Thanks – but the fix was already present in the release version (0.2.0), so an upgrade to the beta is probably unwarranted.

I'm still looking into this. Getting closer to the root cause. As noted, `react-refresh-webpack-plugin` assumes a window context is present, which is not true for web workers. The fix released by @pmmmwh a few days ago (0.2) fixes _half_ the problem by prefixing a 'safeThis' to the `$RefreshSetup$` object, with `safeThis` being derived from the context. It works for simple cases, but fails for us because we're importing lodash as a module into the algorithm worker. On a dev environment that means we're importing a hot-reloadable version of it, which contains code that refers to `$REFRESH_REG$` on the global context. So it fails as described above.

At this point I'm not sure how to fix it. Maybe some webpack magic, possibly a PR to react-refresh-webpack-plugin, or maybe load the webworker code via a generated blob.

abrie: @ivan-aksamentov Ok, now the build works again. The root cause remains elusive. The latest fix [is this slight tweak](https://github.com/neherlab/covid19_scenarios/blob/cbb7d99bb2bfa2c082446e78e9615bbed06776ce/config/webpack/webpack.client.babel.ts#L191) to the webpack config. With that fix the algorithm webworker module loads the non-HMR version of lodash and everything seems to work fine. I do not fully understand this fix, but have not detected anything broken. Will look further when time allows.

pmmmwh: Hi - saw that I was mentioned and the error in your build seems to relate to the webpack plugin. If possible, mind filing an issue in the plugin's repo with proper stack trace/which file the error is pointing to? The `$RefreshReg$` variable is usually injected by the Babel plugin and referenced by the plugin only, so if you have some libraries that needed to be used in a worker context, it would be better to not pass it through `react-refresh/babel`. This unfortunately isn't something I can fix.

abrie: @pmmmwh Indeed, the problem stemmed from a worker dependency being passed through babel. Thank you!

config/webpack/webpack.client.babel.ts
```diff
@@ -36,6 +36,8 @@ import { getenv } from '../../lib/getenv'
 
 import babelConfig from '../../babel.config'
 
+const WorkerPlugin = require('worker-plugin')
```
abrie, line 39: Required for bundling webworker javascript modules.

config/webpack/webpack.client.babel.ts
```diff
@@ -136,6 +138,7 @@ export default {
   entry: path.join(moduleRoot, `src/index.polyfilled.ts`),
 
   output: {
+    globalObject: 'self',
```
abrie, line 141: Webworkers have now window context, so we need to declare the self context here in order for hot reloading (HMR/react-refresh) to work correctly.

config/webpack/webpack.client.babel.ts
```diff
@@ -185,7 +188,7 @@ export default {
           'delay',
           'immer',
           'immer',
-          'lodash',
+          // 'lodash',
```
abrie, line 191: `lodash` is used by the algorithm, and the algorithm is imported into a webworker. If this line left uncommented, lodash is modified by babel for hot reloading, which results in webworker throwing an error. This fix needs a better implementation.

src/components/Main/Main.tsx
```diff
@@ -9,7 +9,7 @@ import { Col, Row } from 'reactstrap'
 
 import { AllParams, EmpiricalData, Severity } from '../../algorithms/types/Param.types'
 import { AlgorithmResult } from '../../algorithms/types/Result.types'
-import { run } from '../../algorithms/run'
+import { run } from '../../workers/algorithm'
```
abrie, line 12: Aim to keep the 'run' import signature the same as before the webworker implementation.

src/components/Main/Main.tsx
```diff
@@ -41,19 +41,28 @@ async function runSimulation(
...
+    const newResult = await run(paramsFlat, severity, scenarioState.ageDistribution)
+    setResult(newResult)
+    caseCounts.sort((a, b) => (a.time > b.time ? 1 : -1))
+    setEmpiricalCases(caseCounts)
+  } catch (error) {
+    // Rejected promises are thrown by await functions.
+    // Catch and log the error message to console.
+    console.error(error)
+  }
+  setIsRunning(false)
```
abrie, line 65: This code is essentially the same as before, except that the `run` call is being transparently executed on a webworker. The try/catch block is necessary to catch any errors thrown by the `await`.  The `run` function isn't expected to throw errors, but the webworker might – such as warnings about structured clone failures or exceptions in the worker code.

src/workers/algorithm/index.ts
```diff
@@ -0,0 +1,28 @@
...
+    })
+
+    worker.addEventListener('error', (error) => {
+      reject(error)
+      console.error('Worker error:', error)
+    })
+
+    worker.postMessage(args)
+  })
+}
```
abrie, line 28: Wraps webworker interaction into a promise suitable for `await`. Initialize the worker, add a message handler, start the algorithm, and `resolve` the result. Algorithm exceptions and worker errors are propagated via `reject`.

src/workers/algorithm/worker.ts
```diff
@@ -0,0 +1,13 @@
...
+import 'regenerator-runtime'
+import { run } from '../../algorithms/run'
+
+self.addEventListener('message', (event: MessageEvent) => {
+  const args = event.data
+
+  run(...args)
+    .then((result) => self.postMessage({ result }))
+    .catch((error) => self.postMessage({ error: `Algorithm error:${error.name}:${error.message}` }))
+})
```
abrie, line 13: Here is the actual webworker code. It's a very thin wrapper around the run function.

config/webpack/webpack.client.babel.ts
ivan-aksamentov: @abrie This stuff you are pointing to is only needed in prod, for better tree shaking and in case the package does not ship an es5 version, in order to support old browsers. But hot reloading is dev only, so I think we can make it work by conditioning additional transpilation to prod only. Should also make dev builds a bit faster.

config/webpack/webpack.client.babel.ts
ivan-aksamentov: @abrie Can we change to `import`, for consistency?

config/webpack/webpack.client.babel.ts
abrie: @ivan-aksamentov is that a task for a new PR? It's not immediately obvious to me how to add a production-only conditional in there... need some help. :)

config/webpack/webpack.client.babel.ts
abrie: @ivan-aksamentov Yes!

config/webpack/webpack.client.babel.ts
```diff
@@ -21,6 +21,8 @@ import kill from 'tree-kill'
 import webpack from 'webpack'
 import { BundleAnalyzerPlugin } from 'webpack-bundle-analyzer'
 
+import WorkerPlugin from 'worker-plugin'
```
abrie, line 24: Use `worker-plugin` to bundle webworker javascript as a module.

src/workers/algorithm/worker.ts
abrie: @ivan-aksamentov At this time, Firefox is unable to transfer Error objects via the structured clone algorithm. So this needs to be a string for now. I'll add a comment to that effect.

https://bugzilla.mozilla.org/show_bug.cgi?id=1556604

config/webpack/webpack.client.babel.ts
ivan-aksamentov: @abrie I added this to master in  #600, please rebase/merge.

abrie: @ivan-aksamentov works great!

ivan-aksamentov:
@abrie  This looks really good! let me know if it's ready for merging.

abrie:
@ivan-aksamentov Great! I feel that's it ready, but this is a pretty significant change – maybe we can get @nnoll and @rneher to give it a run before merging?

ivan-aksamentov:
I hereby summon the masterminds of the project, notably: @rneher @nnoll @noleti ! :boom: 

This is a very important pull request that moves the algorithm execution into a [webworker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API). Webworker is a separate script that runs in a separate context  (think "process") that can only talk to the main thread using [`postMesage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) API (think "IPC").

The advantages is that we can now unlock the main thread and draw in it while the algorithm is running, making UI less laggy and allowing for things like progress indicator (in the button currently)

The disadvantages are that this is the whole another can of worms, with cryptic concurrency bugs, events that don't support transferring some of the objects and a peculiar build process.

Browser support currently is very good and I checked that it even runs in IE10.

Please review, verify that we haven't forget anything and tell if you approve this.

ivan-aksamentov:
For further consideration we may consider an (arguably crazy) idea of aggregating webworkers into a thread pool, for example with [`threads`](https://threads.js.org/usage-pool) package and running instances of the algorithm in this pool, instead of sequential looping (mapping to be precise):

https://github.com/neherlab/covid19_scenarios/blob/cbd67765ed876683d44f864a093d0956c8821693/src/algorithms/run.ts#L31-L45

noleti:
Well, it works in my setup (firefox 75 on Ubuntu 18.04) and Android 10. As .js is not my area, I won't be able to say much on the code. Regarding the webworker communication: I know there are some security pitfalls when it comes to `postMessage()`. A starter is given here: https://gist.github.com/jedp/3005816 . I don't see origin-related things in the current code, but maybe I just missed them.

Overall I think it is good to split the UI and simulation things into different workers. Maybe this will also help with providing "no UI" APIs later (if the sim workers can be run headless).

rneher:
This looks very good to me! I think the visual indication that the calculation is running is super important. And this will allow us to up the number of stochastic sims without producing lags and confusing the user. It runs well on my set-up.

ivan-aksamentov:
@abrie Thanks for this excellent work. Let's see how it plays on master.

Issue #504, opened 2020-04-11
Export all data


rneher:
## 🙋 Feature Request
Currently our table export only includes the totals, not age specific numbers. Users have requested these for downstream reporting.

## 🔦 Context
give users access to more data

## 😯 Describe the feature
this could either be a much later table, or many tables (one for each age-group and the totals)

## 💁 Possible Solution
One could generalize the tsv export function. 

## Related
https://github.com/neherlab/covid19_scenarios/discussions/18#discussioncomment-4055

nnoll:
I solved this somewhat by introducing another export button that outputs all stratified data to a tsv. I think this solves most things people would want. The only additional thing would be if people wanted to output subsets of ages but this seems rather esoteric to me.

ccpf:
It would also be helpful if the export included the actual data as well, plus the mitigation periods, not just the model data. I.e., the export should include all the data necessary to recreate the figure.

nnoll:
Hey @ccpf thanks for the feedback! The mitigation periods are included within the exported json. Furthermore, all data we use can be found on this repo here within the data/ folder.

PR #508, opened 2020-04-12
Allow for export of all stratified simulation data

nnoll:
## Related issues and PRs

Solves #504.

## Description

Generalizes the exportSimulation function to accept a variable number of arguments that are keys into userResult. If none are given, it will default to only reporting the total number for each category.

Will add an additional export category that downloads a large tsv table with all age stratified data.

model.ts
utils/exportResult.ts

## Testing

Download tsv files and see that they are properly formatted.

### DIFF
--- a/src/algorithms/model.ts
+++ b/src/algorithms/model.ts
@@ -556,11 +556,14 @@ export function collectTotals(trajectory: SimulationTimePoint[], ages: string[])
   return res
 }
 
-export function exportSimulation(result: UserResult) {
+export function exportSimulation(result: UserResult, ...ageGroups: string[]) {
   // Store parameter values
 
   // Down sample trajectory to once a day.
   // TODO: Make the down sampling interval a parameter
+  if (ageGroups.length == 0) {
+    ageGroups = ['total']
+  }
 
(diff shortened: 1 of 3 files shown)


src/algorithms/model.ts
```diff
@@ -556,19 +556,33 @@ export function collectTotals(trajectory: SimulationTimePoint[], ages: string[])
   return res
 }
 
-export function exportSimulation(result: UserResult) {
+export function exportSimulation(result: UserResult, ...ageGroups: string[]) {
```
ivan-aksamentov, line 559: It seem that the variable arguments are not needed here, as you are destructuring the array on call site anyway.
To make it more robust and less surprising, could we just pass `ageGroups` as a simple array? Perhaps defaulting to an empty array, if needed.

rneher: I also don't really see why the `...ageGroups` is necessary. Why doesn't it work w/o `...`?

nnoll: The ... allows one to call the function like exportSimulation(result, "total", "80+"). Agreed its not necessary but it is more flexible. I've made the changes.

PR #509, opened 2020-04-12
Variable number of new empirical cases now depends on infectious window.

nnoll:
## Related issues and PRs

Solves Issue #154.

## Description

As of now we have hardcoded in the assumption that the infectious window is 3 days in our display of new cases. This is suboptimal in the event that one wants to explore scenarios in which this differs from our default assumption. This PR solves this by parameterizing this computation by our input parameter. 

I've also kept in general insofar that I don't assume this is an integer as the infectious window we give is a average of a distribution so could in principle be a real number.

ChartCommon.ts
DeterminsticLinePlot.tsx
LinePlotTooltip.tsx

## Testing

Check that the function behaves as expected on changing the infectious period.

### DIFF
--- a/src/components/Main/Results/DeterministicLinePlot.tsx
+++ b/src/components/Main/Results/DeterministicLinePlot.tsx
@@ -197,7 +197,7 @@ export function DeterministicLinePlot({
   const tMin = _.minBy(plotData, 'time')!.time // eslint-disable-line @typescript-eslint/no-non-null-assertion
   const tMax = _.maxBy(plotData, 'time')!.time // eslint-disable-line @typescript-eslint/no-non-null-assertion
 
-  const reducedObservationsToPlot = translatePlots(t, observationsToPlot).filter((itemToPlot) => {
+  const reducedObservationsToPlot = translatePlots(t, observationsToPlot(caseBaseStep)).filter((itemToPlot) => {
     if (observations.length !== 0) {
       if (countObservations.cases && itemToPlot.key === DATA_POINTS.ObservedCases) {
         return true
@@ -320,6 +320,7 @@ export function DeterministicLinePlot({
                     <LinePlotTooltip
                       valueFormatter={tooltipValueFormatter}
                       itemsToDisplay={tooltipItemsToDisplay}
+                      deltaCaseDays={caseBaseStep}
                       {...props}
                     />
                   )}
--- a/src/components/Main/Results/DeterministicLinePlot.tsx
+++ b/src/components/Main/Results/DeterministicLinePlot.tsx
@@ -92,21 +92,40 @@ export function DeterministicLinePlot({
 
   const nonEmptyCaseCounts = caseCounts?.filter((d) => d.cases || d.deaths || d.icu || d.hospitalized)
 
-  const caseStep = 3
+  const caseBaseStep = Math.floor(params.epidemiological.infectiousPeriod)
+  const caseDelta = params.epidemiological.infectiousPeriod - caseBaseStep
+
   // this currently relies on there being data for every day. This should be
   // the case given how the data are parsed, but would be good to put in a check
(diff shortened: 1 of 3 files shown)


src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -92,21 +92,40 @@ export function DeterministicLinePlot({
...
+            cc[i - caseBaseStep].cases !== null
+          ) {
+            return verifyPositive(
+              (cc[i].cases as number) -
+                ((1 - caseDelta) * (cc[i - caseBaseStep].cases as number) +
+                  caseDelta * (cc[i - caseBaseStep + 1].cases as number)),
+            )
+          }
+          return undefined
+        }
```
ivan-aksamentov, line 122: I don't know how to read this. Can we please agree on limiting ternary operators to one-liners only?

As there are 4 booleans combined in an if statement, it makes me think that this particular check does something very specific. But I cannot immediately see what. Can we refactor it into a named function, so that the name would give me a hint?

Please refactor all duplicated statements like `i - caseBaseStep` and `cc[i - caseBaseStep]` into well-named variables.

Why type-casting (`as number`) is needed? I am trying to introduce type checks into the codebase and casts go against that. After cast is written it is extremely hard to find them. I'd prefer you leave type errors that I can see and fix later, or even better that you fix them right away.

Do we really need to produce 2 arrow functions conditionally? Can this condition be moved inside the function?
Can then this function be a normal named `function` in the outside scope, so that the component contains less fluff?

src/components/Main/Results/DeterministicLinePlot.tsx
```diff
@@ -92,21 +92,40 @@ export function DeterministicLinePlot({
...
 
   const countObservations = {
     cases: nonEmptyCaseCounts?.filter((d) => d.cases).length ?? 0,
     ICU: nonEmptyCaseCounts?.filter((d) => d.icu).length ?? 0,
     observedDeaths: nonEmptyCaseCounts?.filter((d) => d.deaths).length ?? 0,
-    newCases: nonEmptyCaseCounts?.filter((d, i) => newCases(nonEmptyCaseCounts, i)).length ?? 0,
+    newCases: nonEmptyCaseCounts?.filter((_, i) => newCases(nonEmptyCaseCounts, i)).length ?? 0,
```
ivan-aksamentov, line 128: Be careful with `_`. You may accidentally stumble upon `lodash`.
JS convention is to put `_0, _1, _2` for unused params. And more often than not, it is a bad sign. Here for example, can't we just pass the candidate value directly, instead of retrieving it by index inside the function? Would probably make the function cleaner as well.

src/components/Main/Results/DeterministicLinePlot.tsx
rneher: I agree this is too convoluted. But it should be pretty straightforward to break into more explicit code. 

One other thing: shouldn't `i - caseBaseStep + 1` be `i - caseBaseStep - 1`?

rneher:
Overall this is a useful improvement that aligns the displayed `new cases` with the infectious curve of the model. Pending the requested changes, I approve of this.

src/components/Main/Results/DeterministicLinePlot.tsx
nnoll: In my opinion producing one or the other function in one branch statement is cleaner than checking this everytime the function is called within the body. But I've made the required changes. The code should be much more obvious now

src/components/Main/Results/DeterministicLinePlot.tsx
nnoll: I've changed this now. It cannot take the candidate value because the very nature of the computation requires it to know past values. It needs the index, not the data.

Issue #519, opened 2020-04-12
PopulationData/ Scenario improvements


noleti:
# 🙋 Feature Request

- Increase the number of countries that we provide scenarios for
- Use public data sources as far as possible to populate populationData.tsv
- This should allow to generate scenarios for more countries?

## 🔦 Context

- We have measured data for many countries, but not many scenarios
- Bottleneck might be data in populationData?
  - ICU data in particular is hard to get
  - Many entries in current populationData have no source references
- Some entries in current populationData are wrong (i.e. name of United Kingdom should be `United Kingdom of Great Britain and Northern Ireland`) based on country_codes.csv

## 😯 Describe the feature

- Generate a script to pull and merge official sources to build populationData.tsv
- Script should incorporate old values of populationData.tsv if available
- Script should attach appropriate source strings
- If/ when we find more data sources, we just need to integrate into script
- If we want to estimate beds or ICU beds based on countries where data is not available, this would be easy and transparent in this script

## 💻 Examples

- If data is available, appropriate entries in populationData.tsv will be created with source strings
- If only partial data is available, we could estimate the rest, and note so in the source strings automatically

## 💁 Possible Solution

- Will create a related PR soon

## Related

  - Related to https://github.com/neherlab/covid19_scenarios/pull/497, https://github.com/neherlab/covid19_scenarios/issues/350, https://github.com/neherlab/covid19_scenarios/issues/341, https://github.com/neherlab/covid19_scenarios/issues/345

noleti:
Currently, entries generated for populationData.tsv can have empty values for population, hospitalBeds, or ICUbeds. Not sure if this is tolerated, or entries like that are useful at all. We could also automatically estimate number of hospital or ICU beds. Might be better than nothing? 
What do the 'scenario-people' say to this?

noleti:
Note that the PR currently only adds 3 additional complete entries to populationData.csv. 16 more entries could be added if ICU beds are estimated based on hospital beds.

PR #520, opened 2020-04-12
PopulationData cleanup, extension, semi-automation

noleti:
Script to build populationData based on five sources, and old populationData.tsv

## Related issues and PRs

- Contributes to https://github.com/neherlab/covid19_scenarios/issues/519

## Description

- This PR introduces a script that uses five official sources to collect data on the population, hospital beds, and ICU beds to generate entries in the populationData.tsv
- In addition, the script integrates data from the old populationData.tsv

- This script automatically adds source string for numbers pulled from these 5 sources
- old source strings for data that could not be pulled automatically is reused together with the old data

- generation of populationData.tsv, which is used in scenario generation AFAIK

## Testing

- Compare generated .tsv with old populationData.tsv


data/hospital-data/ICU_capacity.tsv
```diff
@@ -30,4 +30,3 @@ Spain 	124194 	269 	  	  	4479 	9.7 	3.6 	314.8
 Sweden 	26131 	278 	  	  	550 	5.8 	2.1 	834.0
 Switzerland 	28096 	357 	  	  	1400 	11.0 	3.1 	609.6
 United Kingdom of Great Britain and Northern Ireland 	147809 	237 	1737 	2377 	4114 	6.6 	2.8 	547.0
-CH-Basel-Stadt	1000				80
```
noleti, line 33: This must have been a manual legacy entry, as we use three letter codes now

noleti:
This is ready for comments/reviews now. `make population` will now re-create the `data/populationData.tsv`. As the data should not change often, this will not be required regularly. One advantage of this explicit script is that we now know where data is coming from (which I find quite important). 

Scenario creation via `generate_data` works from what I can tell. I currently assume there are age distributions for all countries, but I was not able to verify this. Can someone double-check?

PR #531, opened 2020-04-13
#443/Allow fine grained values in numeric inputs while preserving step size.

### LINKED ISSUE #443: Allow for fine-grained values in numeric inputs while also preserving step size
# 🐛 Bug Report

Some numeric boxes:
 - have step size defined (a value by which value will be incremented/decremented when clicling arrows)
 - allow for finer values to be valid

For example, for the mitigation strength, we want arrows to increment by 0.1, but we allow values like 0.48 too. Currently this does not work, you can have either one or another:

Screenshot from 2020-04-06 22-46-51

## How to reproduce

Steps to reproduce the issue:

 2. Increment-decrement mitigation strength at will
 3. Try to input 0.48, observe the built--in browser validation error

Note, this issue is applied to all of the numeric fields, not only mitigation strength.

## 😯 Current Behavior

Fine-grained values are not accepted

## 🤔 Expected Behavior

Both, step and fin-grained values work

## 💁 Possible Solution

This is related to the validation attributes in `<input />` tags

There are many libraries exist for numeric inputs, however we would prefer a solution that:
 - has all the features that are currently present
 - uses `<input />`
 - looks native for the browser and platform
 - allows for keyboard navigation (e.g. tab focus)
 - accessible for screen readers (not a "build everything from divs" or at least with aria attributes)
 - accessible for navigation
 - can be internationalized

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Browser          |
| Operating System |
|                  |

## Related

 -

abrie:
## Related issues and PRs

Fixes #443, and possibly makes #454 look silly.
Similar to hotfix #534  

## Description

Browsers reject `<input type="number" step="x"/>` as invalid when the value is not a multiple of the `step` attribute. For this app, such behaviour is undesirable because users may want to enter a number with higher precision than permitted by the step size. And, validation is already being performed by Formik. Therefore browser-level validation is not necessary. This PR disables browser level form validation.

This fix is aimed at numeric inputs but in effect applies to all form inputs.

## Testing

1. Select, for example, a mitigation interval value.
2. Change the value by adding a decimal component, like "40.4"
3. **Press enter**.
  - Observe no error message.
4. Test other non numeric fields
 - Confirm behavior remains as expected.

### DIFF
--- a/src/components/Main/Main.tsx
+++ b/src/components/Main/Main.tsx
@@ -188,7 +188,7 @@ function Main() {
             const canRun = isValid && severityTableIsValid(severity)
 
             return (
-              <Form className="form">
+              <Form className="form" noValidate>
                 <Row>
                   <Col lg={4} xl={6} className="py-1">
                     <ScenarioCard


abrie:
Maybe closed prematurely. Reopening.

abrie:
Duplicates #534

PR #532, opened 2020-04-13
use 'undefined' for empirical cases prior to the difference window

rneher:
## Description
The way empirical cases were added to plot resulted in them being registered at the first date of the interval, rather than the last date (as it should since they are called in new cases in the last three days, not next 3 days). This adds undefined to the array for the first 3 days. 

plot

## Testing
double check plots.


nnoll:
Why did deployment fail? Any reason not to merge?

PR #537, opened 2020-04-14
Model uncertainty in input parameters

ivan-aksamentov:
## Related issues and PRs

Resubmit of #479, which was:
An updated version of PR #375 but now living in the validated world of YAML. Discussion #269.

## Description

Run has been changed to iterate over this array and computes an average of both the linear and quadratic terms over all stochastic realizations to allow for easy return of the mean and variance of outcomes.

## Testing
Tests will need to be written for the affected areas. I'll start this as the PR solidifies.

## Requested help

The new form found at src/components/Form/FormRangeSlider is my sketch at the final widget. I view it as a prototype that works on my screen (there are hard-coded values here). My burgeoning React skills are at their limit to make this a robust component that we could use elsewhere potentially, e.g. within the mitigation values. I would like a pro to look at this.


nnoll:
The number of comments in this PR is proportional to its bloat :)

ivan-aksamentov:
@nnoll I would say it's proportional to the number of automatically discoverable defects in code. With a coefficient of 1 :)

Issue #512, opened 2020-04-12
Fix remaining type issues


ivan-aksamentov:
# 🐛 Bug Report

Not that we are eslint- and tslint-clean, type checking is our next priority.

`yarn tsc:watch`

Me and NIck should be able to do that.

## How to reproduce

Steps to reproduce the issue:

 1. Open the application in a browser
 2.
 3.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Browser          |
| Operating System |
|                  |

## Related

 -

ivan-aksamentov:
This was resolved in PRs listed above

PR #539, opened 2020-04-14
Fixed tsc errors in deterministic line plot

ashwinGokhale:
## Related issues and PRs

 - Fixes some tsc bugs in #512

## Description

To fix tsc errors in the deterministic line plot

- ChartCommon.ts
- DeterministicLinePlot.tsx
- chartHelper.ts


ivan-aksamentov:
@ashwinGokhale Thanks for the efforts.

I don't consider legend changes to be a proper fix. In the end it's just a very complex cast.
The root cause that had to be fixed was that the we use recharts v2 beta, while typings were for 1.8.x.


I untangled the logic in the formatter function and updated the typings:
https://github.com/neherlab/covid19_scenarios/commit/9ea736c68af78d36840d342e2c9dfbcf1ff8bfff

The rest of the changes are unfortunately obsolete with the release 1.1 and merging of the new updated model, because the plot will be changed soon too.

Issue #488, opened 2020-04-09
Additional details for PDF print


noleti:
# 🙋 Feature Request

- In general, make PDF printout more self-explanatory
- Include the help text for variables from the app into the PDF printout
- Choose better names for variables than the  internal ones currently used (cases, populationServed, suspectedCasesToday, overflowSeverity?)  
- Give better context of Interventions
## 🔦 Context

While I like the first PDF print feature that was introduced in 1.0.0, I think there are several ways how it could be improved, in particular to serve as "stand-alone document for busy executives". Some of the variables are not self explanatory, and I believe the help text from the app could be provided along with them. 

In addition, some variables such as suspectedCasesToday should be named better in the printout (this one is actually the number of cases on day 1 of the plot, IMHO).

There could be automated explanations on the meaning of infectiousPeriod and latencyTime, R0, etc. Possibly in addition to the tables already present, but in human readable sentences (see below for examples).
## 💻 Examples

- Instead of just 'ICUBeds: 80', it could be 'ICUBeds: 80. Note: Number of ICU/ICMUs available in health care system. Presets are rough estimates indicating total capacity. Number of ICU/ICMUs available for COVID-19 treatment is likely much lower.' (or something slightly shorter)

In the following, <foo> is a placeholder for the variable value.

- Instead of 'suspectedCasesToday: 39', put 'initially infected on <day1> of simulation'.

- 'Our model assumes that infected people take on average <latencyTime> days to show symptoms and become infectious, and will recover after <infectiousPeriod>, unless their condition degrades to severe, at which point they will stay in a hospital for <lengthHospitalStay> days. If they need to be transferred to the ICU, they will require care for <lengthICUStay> days on average. While people are infectious, they infect <R0> other patients on average in absence of mitigations.'

## 💁 Possible Solution

Feature should require templates in .js code to prepare the print page. Maybe some templating engine with separate templates?

## Related

 - https://github.com/neherlab/covid19_scenarios/pull/471
 -

rneher:
some of these issues were addressed in #489, but there are still aspects that can be improved

rneher:
adding this to milestone 1.1. I think some proof-reading, a few words here and there will go a long way.

ivan-aksamentov:
I am keeping this open as a reminder. This issue is meant for a domain expert or a scientist. Here we want to restructure and expand the PDF in a way that makes it more useful for reading by another domain expert or a scientist.

[...]

PR #540, opened 2020-04-14
minor edits to print page

rneher:
## Related issues and PRs
#488 

## Description
Increase readibility and information content of the print page. 

print page

## Testing
render print page.


rneher:
we could replace the parameter lists by tables similar to the results table.

PR #549, opened 2020-04-15
Okinawa

spigolotti:
## Description
Include parser for Okinawa data and model parameters for the Okinawa scenario

all changes are in the data folder (parser, sources, fit parameters, initial conditions)


rneher:
thanks! But this PR should only contain the parser and the additions to `populationData.tsv`. The update of the data will only generate merge conflicts.

PR #550, opened 2020-04-15
Added Philippines and Philippines-regions

jilladona:
## Related issues and PRs

 -


noleti:
Hi @jilladona , thanks for this! Numbers look good, and match what I found elsewhere for the country as well. As a minor note, I was not able to reproduce your computation of ICU beds. Your note says that they are derived from hospital beds, but there is no fixed ratio I could see between the states.

jilladona:
Hi, @noleti! Here's the worksheet I used for ICU beds per region. It takes the beds and hospitals counts from the EU report on Philippines Health System. Levels 2 and 3 hospitals have ICUs, but only one region has data on number of hospitals by level, so I apply these ratios to the rest of the regions. It will not add up to the number of Philippines ICU beds, which is from a different data source (2020). But we think this is a good estimate per region.

[PHL hospital capacity.xlsx](https://github.com/neherlab/covid19_scenarios/files/4480024/PHL.hospital.capacity.xlsx)

noleti:
Thanks for the details, this looks good to me. @rneher should I merge, or will you?

PR #551, opened 2020-04-15
added Okinawa parser and Okinawa population data

spigolotti:
## Related issues and PRs

 -


spigolotti:
I created a new fork with only the Okinawa parser and the population data. Please disregard the previous pull request.

noleti:
Hi @spigolotti , the entry in the `sources.json` is also required for a parser, but that should be it. Can you add it to this PR?

rneher:
thanks. @noleti would be great if you had a look. We hopefully have a unified source for all of Japan soon.

rneher:
One more thing: the csv doesn't have data for every day. this might cause problems downstream (We should actually handle this better). quick fix would be to simply fill in the days in the csv, or expand the in the parser.

spigolotti:
I added the data in the sources.json

spigolotti:
as for the missing days, I'll fill in the csv, it is probably easier

noleti:
@spigolotti  Do you have a source/comment for the hospital and ICU bed number? I checked with my data on Japan in general, and for all of Japan there are roughly 0.000073035 ICU beds per inhabitant. Your numbers are 4 times lower. Your hospital bed number is 10 times lower than expected (not that this matters for simulation currently).

spigolotti:
A source for the number of ICU beds is https://www.stopcovid19.jp . I could not find information about other beds available for COVID-19 patients - I understand that Okinawa did not have beds other than ICUs for an infective disease such as COVID-19, and now they are organizing new beds. The number 100 is just a rough estimate of the number of beds they might be able to set up. I could not  find more precise information for this latter number (and as you say it does not really matter for simulation purposes).

noleti:
@spigolotti Thanks, this looks like good enough data for me.

data/populationData.tsv
```diff
@@ -240,3 +240,5 @@ FRA-Martinique	364354	France	882	20	Tropical	None	None	None
 FRA-Mayotte	270372	France	300	15	Southern	None	None	None
 Israel	9136000	Israel	15871	758	Northern	None	None	None
 Iceland	364260	Iceland	1169	29	Northern	None	None	None
+JPN-Okinawa	1433566	Japan	100	24	Tropical	None	None	None
+
```
noleti, line 244: This empty line currently breaks a script downstream, please remove it. While you are at it, can you put the website you mentioned as src for the ICU bed number (last column?)

data/parsers/okinawa.py
```diff
@@ -0,0 +1,43 @@
...
+# ------------------------------------------------------------------------
+# Main point of entry
+
+def parse():
+    r  = requests.get(URL)
+    if not r.ok:
+        print(f"Failed to fetch {URL}", file=sys.stderr)
+        exit(1)
+        r.close()
+#    regions = dict(JPN-Okinawa=[])
```
noleti, line 25: `regions = {'JPN-Okinawa'=[]}` would have been the right way ;) Can you remove the commented out old code in this parser?

spigolotti: removed empty line, added the ICU source and removed the commented old code

PR #552, opened 2020-04-15
add R0 to the mitigation plot

rneher:
## Description
Plot the resulting R0 along with the uncertainty and mitigations. 

Need to sort out the joint display of mitigations and R0

types, plots, run


rneher:
I think overall this is an improvment. @nnoll have a look!

nnoll:
@rneher this looks great and solves the problem I was having last night of how to show both the intervals of mitigation and the uncertainty in R0. We can probably think of better labeling for each plot. I'll merge this in if you are done tinkering.

PR #536, opened 2020-04-13
Update initial values for Ukraine

Vanuan:
https://web.archive.org/web/20200328082145/https://moz.gov.ua/article/news/operativna-informacija-pro-poshirennja-koronavirusnoi-infekcii-2019-ncov-

PR #556, opened 2020-04-15
Automate selection of case thresholds for data fitting procedure

nnoll:
## Related issues and PRs

 No issue has been opened yet, but this is because data is fit so infrequently. This will be a problem upon subsequent fits.

## Description

Last night I tried to rerun the fits to include new data that had been contributed. I ran into a problem in that our estimates for many countries, including our default scenario the US, changed dramatically. The reason for this was simple, we were fitting cases filtered on a crude cutoff. Thus, for example with the US, we were fitting way too deep into the epidemic and thus were picking up mitigation efforts.

This will likely be fixed when/if we estimate mitigation efforts, but in the interim I thought we needed a better solution here. My solution is to quickly fit a logistic function to the data. The maximum case counts is chosen to be 1 doubling after the midpoint of the logistic function and the minimum case counts is chosen to be 10 doublings before. 

model.py

## Testing

Please look at many different scenarios to see if the initial values are sensible.


nnoll:
This also supercedes PR #536.

PR #559, opened 2020-04-16
Feat: better PDF layout

ivan-aksamentov:
## Related issues and PRs

 - Contributes to #110
 - Extends #471 

## Description

 - [x] cleanup of the PDF preview code
 - [x] all links are now blue (in the entire app)
 - [x] better page layout (still not perfect, because the components towards the end are ugly by themselves)
 - [x] large paragraphs are now in mdx, so that they can be edited easier. It is desirable to add some more text in each section, with new mdx components (cc @rneher )
 - [x] logos
 - [ ] names, affiliations, emails?
 - [ ] preview looks like a page (it is not), but "back" button wouldn't work as expected. Not sure what to do about it

print preview and pages, links 

## Testing

Manual, visual.

IMPORTANT: thoroughly test in Chromium and Firefox. They both have quirks. Test both A4 and Letter paper format.


noleti:
I'm attaching my output from Firefox 75.0 on Linux. 
[covid.pdf](https://github.com/neherlab/covid19_scenarios/files/4484955/covid.pdf)

In general, this looks nice. Ideally, the 

- For mitigations: Reduction of transmission could be explicitly in % (only 50 given right now)
- Can text links be turned into working hyperlinks? For people reading the PDF on a tablet or computer. There could even be a clickable link to the URL that opens the specific scenario (don't print the full link, just embed it as clickable).
- Results table does not list overflow. This means that in case of overflow, the ICU number will be stuck at ICUBeds for a while without direct feedback that there is a big problem going on. I think there is horizontal space for another column?

rneher:
A big improvement! two thoughts:
 
 * we previously had the age distribution labeled with small numbers. that might still be a useful. 
 * we should add overflow ICU to the table. Currently ICU maxes out the capacity and the overflow (most important for planning) isn't shown.

ivan-aksamentov:
@noleti @rneher @nnoll 

Update: 

 - [x] fixed size of the plots in pdf preview on mobile as reported by Nick. There are still major problems on mobile. But getting tables to render properly on narrow iPhone and on A4 page is a tough task. Not sayin it's impossible, but probably not at this time.
 - [x] added ICU overflow column to the results table
 - [x] formatted mitigation strength as percent
 - [x] added print labels back on age distribution plot
 - [x] unset body color when printing

PR #583, opened 2020-04-17
Use generated types for age group specific parameter grid

gj262:
## Related issues and PRs

Relates to #283.

## Description

This changes the age group specific parameters editable grid to use externally defined types `Severity` and `AgeDistribution` and not propagate the display oriented `SeverityTableRow` type. This is largely a refactor, simplification and addition of unit test, behavior remains the same.

Age group specific parameter grid.

## Testing

- Change parameters, check for propagation to run() in autorun and manual.
- Enter invalid parameters, check that errors are displayed and that Run is disabled in manual mode and run() does not occur in autorun mode.
- Fix parameters - run should work as before.


gj262:
Codeclimate issues are unrelated.

PR #586, opened 2020-04-17
Add (sum up) regional data to obtain country data for France

Christian-G-UT:
## Description
Add regional data, which is more detailed (ICU, hospitalized, etc), to obtain country data for France

add_cases in utils.py needed change since data for France is irregular, so a check for None values was done
france.py was modified to add regional data

### DIFF
--- a/data/parsers/france.py
+++ b/data/parsers/france.py
@@ -4,7 +4,7 @@ import csv
 import io
 
 from collections import defaultdict
-from .utils import store_data, stoi
+from .utils import store_data, stoi, add_cases
 
 # ------------------------------------------------------------------------
 # Globals
--- a/data/parsers/france.py
+++ b/data/parsers/france.py
@@ -63,5 +63,7 @@ def parse():
     regions2 = {}
     for reg, d in regions.items():
         regions2['-'.join(['FRA',reg])] = [d[day] for day in sorted(d.keys())]
+        
+    regions2 = add_cases(regions2, list(regions2.keys()), 'France', cols)
 
     store_data(regions2, 'france',  cols)
--- a/data/parsers/utils.py
+++ b/data/parsers/utils.py
@@ -115,7 +115,8 @@ def add_cases(cases, toAdd, target, cols=default_cols):
                 if d[cols.index('time')] == time:
                     new = False
                     for i in range(1,len(d)):
-                        d[i] += e[i]
+                        if e[i]!=None and d[i]!=None:
+                            d[i] += e[i]
             if new:
                 # we did not have that date in the aggregate yet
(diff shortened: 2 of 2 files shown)


data/parsers/utils.py
```diff
@@ -115,7 +115,8 @@ def add_cases(cases, toAdd, target, cols=default_cols):
                 if d[cols.index('time')] == time:
                     new = False
                     for i in range(1,len(d)):
-                        d[i] += e[i]
+                        if e[i]!=None and d[i]!=None:
+                            d[i] += e[i]
```
noleti, line 119: Doesn't this lead to unexpected behaviour if d[i]= None? In that case, e[i] would not be used (although it probably should be replacing d[i]). How about 

```
if d[i] != None:
    if e[i]:
        d[i] += e[i]
elif e[i] != None:
    d[i] = e[i]
    
```

data/parsers/france.py
```diff
@@ -63,5 +63,7 @@ def parse():
     regions2 = {}
     for reg, d in regions.items():
         regions2['-'.join(['FRA',reg])] = [d[day] for day in sorted(d.keys())]
+        
+    regions2 = add_cases(regions2, list(regions2.keys()), 'France', cols)
```
noleti, line 67: We get France country data from ECDC parser currently. Did you compare the values? I will have to look into what happens if we have the same country listed twice (I think it depends on the alphabetic order of folders, so right now the ecdc numbers might take precedence in the .json generation).

data/parsers/utils.py
noleti: But in general, I think this is a good observation and definitely should go into this function.

data/parsers/france.py
noleti: You are missing the add_cases import in this file as well

noleti: I looked into the the merging question now. As I predicted, the ECDC data takes precedence over the france parser currently. That being said, ECDC does not provide hospital and ICU data, so this PR is definitely valueable, as we fill missing ECDC data with output of this parser. As a note, france parser's death count is considerably lower than the ECDC data, and we currently use the ECDC data instead.

Christian-G-UT: Thanks for looking into this and figuring everything out. The reason for the ECDC death data being considerably higher is most probably because it is considering overall Covid deaths while the regional source of the parser only hospital death. Most valuable is indead to have the values of ICU and hospitalized for France as a whole by summing the regional data. Thanks again.

data/parsers/utils.py
Christian-G-UT: I agree, I did the above changes.

Issue #580, opened 2020-04-17
Improve UI and UX around new uncertainty functionality


ivan-aksamentov:
# 🙋 Feature Request

With new parameter (and results) uncertainty functionality we now need to build a new user experience around this features.

## 🔦 Context

After #537 merged we now have a very important algorithmic improvement:

 - instead of specifying the exact value, users now can specify the range in which parameter is distributed (currently R0 and mitigation strengths)
 - the "Results" plots now show mean value (line) and the distribution where the result can fall with some probability (colored areas).
 - numeric representation now shows mean value and upper and lower bound
 - currently input parameters range is represented by a slider

## 😯 Describe the feature

Play with different representations and suggest the most suitable: 
 - [ ]  of parameter input for R0: maybe slider with 2 or 3 knobs, with or without editable numbers?
 - [ ]  of parameter input for mitigation strength: maybe slider with 2 or 3 knobs, with or without editable numbers?
 - [ ] representation of the output results on the plot (min, lower, upper)
 - [ ] representation of the output results in numeric form, such as in the plot tooltip
 - [ ] representation of the output results in the results table in the print preview page

## Related

 - #537 
 - #269

gj262:
For parameter input I'd suggest switching to a freeform input that would allow single values and a range separated by 'to' or '-'. Ideally it would provide a hint that a range can be entered as the user types. A tooltip and placeholder could be used to explain the options.

![Typing hint](https://user-images.githubusercontent.com/6854428/79669614-920f8680-8171-11ea-9bdb-8ebeb76c8f6a.png)

![With a range](https://user-images.githubusercontent.com/6854428/79669615-963ba400-8171-11ea-97aa-b77420dcb9e3.png)

ivan-aksamentov:
This was mostly resolved in #604 and #643

PR #604, opened 2020-04-20
feat: replace range sliders with pairs of spinboxes

ivan-aksamentov:
## Related issues and PRs

 - Contributes to #580 

## Description

This explores the user experience with 
 - a pair of spinboxes as a mean to set ranges of numbers for the parameter uncertainty feature
 - table layout for the mitigation widget

UX

### DIFF
--- a/src/components/Form/FormRangeSpinBox.tsx
+++ b/src/components/Form/FormRangeSpinBox.tsx
@@ -3,23 +3,12 @@ import React from 'react'
 
 import _ from 'lodash'
 
-import { FormikErrors, FormikTouched } from 'formik'
 import { Col, FormGroup, Row } from 'reactstrap'
+
+import type { FormSpinBoxProps } from './FormSpinBox'
 
 import FormLabel from './FormLabel'
 import { RangeSpinBox } from './RangeSpinBox'
-
-export interface FormRangeSpinBoxProps<T> {
-  identifier: string
-  label: string
-  help?: string | React.ReactNode
-  step?: number | string
-  min?: number
-  max?: number
-  pattern?: string
-  errors?: FormikErrors<T>
-  touched?: FormikTouched<T>
-}
 
 export function FormRangeSpinBox<T>({
   identifier,
@@ -31,7 +20,7 @@ export function FormRangeSpinBox<T>({
   pattern,
   errors,
   touched,
-}: FormRangeSpinBoxProps<T>) {
+}: FormSpinBoxProps<T>) {
   const isTouched = _.get(touched, identifier)
   const errorMessages = _.get(errors, identifier) as string[]
   const showError = errorMessages && isTouched
--- a/src/components/Form/FormRangeSpinBox.tsx
+++ b/src/components/Form/FormRangeSpinBox.tsx
@@ -1,13 +1,15 @@
+/* eslint-disable react/no-array-index-key */
 import React from 'react'
 
 import _ from 'lodash'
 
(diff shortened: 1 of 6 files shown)


rneher:
I think this is ultimately easier than the slider. 
 * the allowed ranges should be such that 0<min<max<6 for R0 and 0<min<max<100 for mitigation measures
 * the boxes could be much smaller (with don't expect this to have more than 2-3 digits ever)

rneher:
It looks like @gj262 has also worked on this
https://github.com/gj262/covid19_scenarios/commit/73b1806be86f3b113dacd0682ee4aed732fe9c89

Gavin's solution is actually quite intuitive and allows both single value and range inputs!

gj262:
> It looks like @gj262 has also worked on this
> [gj262@73b1806](https://github.com/gj262/covid19_scenarios/commit/73b1806be86f3b113dacd0682ee4aed732fe9c89)
> 
> Gavin's solution is actually quite intuitive and allows both single value and range inputs!

My change though is still somewhat vaporware. And dealing with long fractional parts e.g.`2.9700000000000006` is easier with dual inputs. (Those long fractional parts are probably bugs but I don't think we want the UI rounding (even though the slider already effectively does that)).

ivan-aksamentov:
@rneher 

 - Is there a particular reason to limit R0 to 6? We did not limit it before, I don't see why limit it now. I typically prefer to avoid "640K ought to be enough for anyone" judgments where possible, because one can never know what are the usecases out there. For example, what if someone wants to show that the values above 6 are not feasible in real life?

 - Boxes are smaller than what you see for most people running on laptop screens. In fact, on some sizes the boxes are too small (barely fitting 2 numbers), so it will probably require some adjustment.

 - Regarding Gavin's solution, it is indeed very pretty, but, as a user, if I would want to get any job done, I would not want that UX: there are no arrow buttons, values have to be typed manually every time. There is a syntactic constraints. For example, if deleted, need to type the " to " keyword in between and/or to precisely aim the mouse cursor to where the number is. It would be an excellent solution for a read-only field, but for the purpose of tweaking and trying different numbers this does not work well in my opinion. Plus it a bit too much of the internal code logic for a very slim gain. Implementing from scratch an input that does not suck is a non-trivial task and there are many examples of failure on npm.

The solution in this PR chooses a way of the least surprise: need to input 2 values - here are the two input boxes, native to the browser. However the semantics of a range still has to be enforced somehow.

For that purpose I was also planning to play with an extended version with adding a slider (so, 2 boxes AND a slider), however I am not sure if I should prioritize another feature instead (e.g. import)

cc @nnoll @noleti what is your opinion on range inputs?

noleti:
My opinion: the spin boxes are better than the sliders (in particular, if we can cut precision to 2 decimals or so). It still seems more effort to use than the old solution for mainstream users. How are the spin boxes planned to be integrated with the auto refresh? A refresh on each cursor button press is maybe too much.

My alternative proposal: One box provides the mean, and the other one the +- delta to the mean. So instead of 2.7 3.3, you would have 3 0.3. The second box could also be the len of the range, i.e. 0.6 in this example. The main intuition is that this allows to shift the mean of R0 easily by modifying one number, instead of having to change 2 all the time. The second input box would essentially configure the range of alternatives to explore.

ivan-aksamentov:
@noleti I was suggesting a `mean ± delta` approach as well, but then scientists said that it looks like a Gaussian. And we draw from uniform. So we kept it as `min - max` for now.

Issue #597, opened 2020-04-19
Provide populationData source strings in app


noleti:
# 🙋 Feature Request

- Provide data source info to users in the app if requestsed
## 🔦 Context

- Allow user to judge quality of data used by us (e.g., if ICU estimated or hard sources)

## 😯 Describe the feature

- Pull src strings from populationData, inject into scenarios.json, use in app UI (e.g., as part of the help string provided for individual variables)

## 💻 Examples

- For Germany, when clicking on question mark for ICU, it could have the normal text, plus "source: Eurostat (https://ec.europa.eu/eurostat/tgm/table.do?tab=table&init=1&plugin=1&language=en&pcode=tps00046)"

## Related

 -
 -

realdy:
Hello, I'm interested in helping out for this feature, but I'm still familiarizing myself with the structure of the repo. Would you mind providing the locations of the files you think would be necessary to add the feature?

noleti:
The strings are stored in `data/populationData.tsv` (for many population constraints) and would likely need to be processed in `data/scripts/scenarios.py` to be included in the generated `src/assets/data/scenarios/scenarios.json`. From there, they would have to be picked up in the app somewhere. I am not involved in the app development myself, so I can't give more hints afterwards. Maybe we can wait for feedback from @ivan-aksamentov to see if this feature is wanted at the current state before proceeding.

realdy:
Okay, thanks for the pointers you've given me so far! It's helpful to have something to work with.

And that makes sense. I could certainly see this feature as being helpful, but there might be a lack of visibility in its current proposed form. It's not immediately obvious to the user that clicking on the ICU/ICMU question mark button would show the user where the data is sourced from. So getting some feedback before making a solid build would be ideal.

That said, I'll still go ahead attempting to make progress since it would be helpful to understand the structure of the repo and I think that including the information somewhere would still be helpful, even if it's not in the originally proposed form.

realdy:
I managed to create a commit that includes the srcHospitalBeds parameter in the generated json file if you would like to review it.

Working on adding the relevant data to the app

realdy:
To get some verification, when you were thinking of the structure of the feature, did you want it to look something like this:
![Example](https://user-images.githubusercontent.com/45107348/79753086-58f72380-82e3-11ea-8ec9-a5b353327fd9.png)

The idea is that whenever a location is selected, the text in the help box  will refresh based off the data in the json file

noleti:
[...]

PR #606, opened 2020-04-20
Provide populationData source strings in app #597, Part1

realdy:
## Related issues and PRs
Related to Issue #597
Changed scenarios.py to include srcHospitalBeds in generating json  

## Description
Review changes to ensure that initial changes for scenarios.py are sufficient and do not cause any severe issues before moving on to expressing the changes within the app

scenarios.py
scenarios.json
scenarios.yml
type.py

## Testing
Run "make all" and verify that there are no errors.
Inspect the generated scenarios.json to verify that srcHospitalBeds is included
Also inspect type.py to verify that "src_hospital_beds" is an include parameter


schemas/Scenarios.yml
```diff
@@ -91,7 +91,11 @@ definitions:
         type: integer
         multipleOf: 1
         minimum: 0
-
+      
+      srcHospitalBeds:
+        type: string
+        minLength: 1
+
```
noleti, line 98: You are currently not setting srcHospitalBeds as required in the schema. This means the app will have to tolerate missing keys. I'm not sure if we should set it as required or not, we will need to see later.

noleti:
Hi @realdy, please don't check in changed in `data/case-counts` and `src/assets/data/case_counts.json`, as these are unrelated to the code changes, and make the PR very hard to review. In particular, the inclusion of newer data leads to many changes in the `scenarios.json`. 

From what I can tell, the other changes look good so far. As I noted, we will have to decide whether the src strings should be required by the schema, or be optional.

schemas/Scenarios.yml
realdy: Yes, I wasn't sure if it was necessary to make it required or not given its tenuous inclusion at the moment, but it's only a 1 line change in code, so we can always account for it later. From my understanding (albeit rudimentary), not making it required wouldn't flag some sort of error in case the parameter isn't included. Given that we aren't deadset on this feature, I figured having that sort of flexibility wouldn't be too detrimental.

And understood. I certainly figured that the tsv's would be rather excessive if included.
To verify, is there some way I can edit my pull request to ensure those aren't included, or should I just be mindful in future commits to not include those files?

noleti: I don't have a local checkout of this branch, so I think you should be safe to delete your last commit via resetting the head and force-pushing the branch. Or you just revert the commit (which keeps it in the history, with another undo-commit). In both cases, you can then make another commit on the branch which changes only the needed files.

realdy: I believe the latest commit was properly reverted and should have only the necessary changes, but let me know if I didn't do it correctly.

src/components/Main/Scenario/ScenarioCardPopulation.tsx
```diff
@@ -21,12 +22,57 @@ countryOptions.push({ value: CUSTOM_COUNTRY_NAME, label: i18next.t(CUSTOM_COUNTR
...
+  srcICUBeds?: string
+  srcPopulation?: string
+  scenarioName: string
 }
-
-function ScenarioCardPopulation({ errors, touched }: ScenarioCardPopulationProps) {
+function getSrcStrings(srcHospitalBeds?: string, srcICUBeds?: string, srcPopulation?: string) {
+  let ret = ['Source: ', 'Source: ', 'Source ']
+  if (srcPopulation === undefined || srcPopulation === 'None') {
+    ret[0] = 'Source cannot be provided'
```
noleti, line 44: Why are the negative messages not all the same (provided vs found?) I would prefer something like 'No data source in our records.' or similar for all of them.

src/components/Main/Scenario/ScenarioCardPopulation.tsx
```diff
@@ -43,7 +89,7 @@ function ScenarioCardPopulation({ errors, touched }: ScenarioCardPopulationProps
       <FormSpinBox
         identifier="population.populationServed"
         label={t('Population')}
-        help={t('Number of people served by health care system.')}
+        help={t('Number of people served by health care system.\n'.concat(srcStr[0]))}
```
noleti, line 92: The '\n' does not seem to work, can you use a space instead?

src/components/Main/Scenario/ScenarioCardPopulation.tsx
```diff
@@ -78,8 +124,12 @@ function ScenarioCardPopulation({ errors, touched }: ScenarioCardPopulationProps
       <FormSpinBox
         identifier="population.hospitalBeds"
         label={`${t('Hospital Beds')} (${t('est.')})`}
+        // There's a bug in rendering special characters. For example for, "https://..."" the ':' character causes an error in rendering
+        // Also need to find out how to input newline characters just for nice spacing
         help={t(
-          'Number of hospital beds available in health care system. Presets are rough estimates indicating total capacity. Number of beds available for COVID-19 treatment is likely much lower.',
+          'Number of hospital beds available in health care system. Number of beds available for COVID-19 treatment is likely much lower.'.concat(
```
noleti, line 130: Can you use a space to separate the previous and new string?

src/components/Main/Scenario/ScenarioCardPopulation.tsx
```diff
@@ -90,7 +140,9 @@ function ScenarioCardPopulation({ errors, touched }: ScenarioCardPopulationProps
         identifier="population.ICUBeds"
         label={`${t('ICU/ICMU')} (${t('est.')})`}
         help={t(
-          'Number of ICU/ICMUs available in health care system. Presets are rough estimates indicating total capacity. Number of ICU/ICMUs available for COVID-19 treatment is likely much lower.',
+          'Number of ICU/ICMUs available in health care system. Number of ICU/ICMUs available for COVID-19 treatment is likely much lower.'.concat(
```
noleti, line 143: Can you use a space to separate the previous and new string?

noleti:
@realdy This looks nice to me, thanks. I made some small comments, but overall this works well for me. 

@ivan-aksamentov I leave this to you to judge if this will complicate the app too much. I find this passing of source information through help text intuitive and less effort than a separate source page (which could still be made). 

We could even have source strings for the age distribution, imports per day (estimated by model.py imho), case-counts, etc (I leave this for the future). 

In general, links in the source strings are not clickable (and would be ideally). We can address that in the future, if this PR is generally accepted.

realdy:
@noleti 
Just made the changes you requested. Thanks for the helpful comments! Still haven't figured out how to deal with the new line character, but if this pr gets accepted I'll go forward with attempting to develop the other features you mentioned. At this point though, I think this is enough for a decent proof of concept and I'd rather not spend any more work on it if the PR ultimately is not accepted.

PR #613, opened 2020-04-21
Change model results to percentiles over geometric mean/standard deviation

nnoll:
## Related issues and PRs

 - 

## Description

The geometric mean and standard deviation led to odd visual results. 
We now will report percentiles over time.

model.ts
run.ts
Changed userResult.trajectory.mean -> userResult.trajectory.middle

## Testing

Check model plots and exports with new data structure.

### DIFF
--- a/src/algorithms/model.ts
+++ b/src/algorithms/model.ts
@@ -396,15 +396,15 @@ export function exportSimulation(result: UserResult, ageGroups: string[] = ['tot
   // TODO: Make the down sampling interval a parameter
 
   const categories = {
-    current: keys(result.mean[0].current),
-    cumulative: keys(result.mean[0].cumulative),
+    current: keys(result.middle[0].current),
+    cumulative: keys(result.middle[0].cumulative),
   }
   const header: string[] = ['time']
 
   categories.current.forEach((category) => {
     ageGroups.forEach((age) => {
       header.push(
-        `${title(category)} (${age}) mean`,
+        `${title(category)} (${age}) median`,
         `${title(category)} (${age}) lower bound`,
         `${title(category)} (${age}) upper bound`,
       )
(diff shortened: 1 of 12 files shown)


nnoll:
@rneher @ivan-aksamentov @noleti: I'm unsure about the notation for median & lower/upper percentiles. Thoughts here?

ivan-aksamentov:
I thought you mentioned a tuple/triplet, for example:

lower, **median**, upper
2.78, **3.14**, 6.28

noleti:
For USA without mitigations, ICU looks good now, but overflow median does not start at 0. I think it didn't before either, I just noticed. Otherwise this looks good. I'm not sure why you call it percentiles here, that just applies if we assume the input range is some distribution, no? The lower, median, upper values plotted are deterministic, no?
icuCap2

nnoll:
@ivan-aksamentov my original thoughts were to do a tuple. I ended up on this after some trial and error but am not wedded to the UX at all.

@noleti the input parameters are ranges now for R0 & mitigation efficacy. You are adjusting the prior distributions (which are assumed uniform) when you play with the slider hence the percentile. I agree that the help buttons text should be updated to reflect this.

nnoll:
@ivan-aksamentov my original thoughts were to do a tuple. I ended up on this after some trial and error but am not wedded to the UX at all.

@noleti the input parameters are ranges now for R0 & mitigation efficacy. You are adjusting the prior distributions (which are assumed uniform) when you play with the slider hence the percentile. I agree that the help buttons text should be updated to reflect this.

rneher:
Ideally one would have something like 
"2020-04-20   (10%, median 90%)"
the title of the tool tip and then numbers as a tuple. The current colon notation is a little odd. I like Ivan's proposal with the bold median.

PR #615, opened 2020-04-21
Round of lower/upper bounds on R0 & mitigation efforts

nnoll:
## Description

Simple fix of the export of scenario growth rates to only keep one significant figure.

scenarios.py

## Testing

Look at the exported data.

### DIFF
--- a/data/scripts/scenarios.py
+++ b/data/scripts/scenarios.py
@@ -107,7 +107,7 @@ class Fitter:
 # Parameter class constructors (with defaults)
 
 def report_errors(x):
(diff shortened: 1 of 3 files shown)


ivan-aksamentov:
Maybe we change rounding to 0.01? This is the step granularity for the R0 form field.

Also, why cases data is changed?

nnoll:
Cases counts update when you run make data-for-app which I used to regenerate the scenarios. They should include additional dates.

nnoll:
Cases counts update when you run make data-for-app which I used to regenerate the scenarios. They should include additional dates.

Issue #517, opened 2020-04-12
Please add model cumulative cases


coderb:
Please add data for the model cumulative cases.

This would be useful when trying to manually fit parameters to observed data especially around interventions.

noleti:
Is this the same as https://github.com/neherlab/covid19_scenarios/issues/504 ?

coderb:
no, it's a request for a new data series which is not currently displayed

noleti:
Can you be more specific? You meant the sum of people who were infected, in hospital, in ICU, or dead? I.e. total population-recovered?

coderb:
the cumulative number of people that have been infected over time, ie the model version of the data series "cumulative cases (data)".

ccpf:
Isn't that: cumulative deaths + icu + severely ill + infectious + recovered? At least this is how I have been interpreting the model output.

noleti:
AFAIK "Cumulative cases (data)" is directly taken from our `data/case-counts/*/*.tsv` files (which are, in turn, taken from third party APIs). So you could directly take the values from our repo. Does that help?

ccpf:
@noleti: true! Thanks for pointing that out. So then coderb is right in that we cannot know the (modelled) cumulative cases? Regarding your answer to him/her, I think the question was about the modelled cumulative cases and not the actual data. Cheers.

coderb:
@noleti i think you misunderstand what i am asking for.  simply to plot (in the graph section) what the model gives for the cumulative number of cases.  i believe the formula would simply be N - S where N is the total population and S is the number of susceptible (assuming there is no change in total population over time).

noleti:
Your N-S is the number I suggested in https://github.com/neherlab/covid19_scenarios/issues/517#issuecomment-612770785. I'm not sure how this will help with manual fitting, as usually there is a big problem with undercounting of infected (but not severe) cases. Fitting is most reliably done based on deaths. 

In any case, I think it is now clear what you are asking for, I will update your initial question to describe it more clearly from my perspective.

coderb:
thank you

nnoll:
Hey @coderb. My worry about including this is as @noleti states, it's probably not wise to fit our cumulative infectious categories directly to the cumulative case counts as there's an unknown underreporting fraction for each scenario. Furthermore, I worry about our UX here: our results plot is already quite dense with information that I fear adding another plot will compound our problem there.

[...]

PR #620, opened 2020-04-21
Feat/cumulative simulation cases

shijurodhaz:
## Related issues and PRs

 - Related to issue #517 

## Description

Add a new line to the results plot to show the cumulative cases in the simulation. This is calculated using population - current susceptible

src/components/Main/Results/DeterministicLinePlot.tsx 
src/components/Main/Results/ChartCommon.ts

## Testing

Check the results plot and compare with population/ susceptibility values

### DIFF
--- a/src/components/Main/Results/ChartCommon.ts
+++ b/src/components/Main/Results/ChartCommon.ts
@@ -22,6 +22,7 @@ export const DATA_POINTS = {
   NewCases: 'newCases',
   HospitalBeds: 'hospitalBeds',
   ICUbeds: 'ICUbeds',
+  CumulativeSim: 'cumulativesim',
   /* Observed */
   ObservedDeaths: 'observedDeaths',
   ObservedCases: 'cases',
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -3131,7 +3131,7 @@ append-buffer@^1.0.2:
   dependencies:
```
ivan-aksamentov, line 3131: @shijurodhaz  Could you please remove any changes in `yarn.lock`?

ivan-aksamentov:
I will need help from our science team here:

@nnoll @rneher @noleti  Do we need this? Initial discussion seems to be stemming from #517

yarn.lock
shijurodhaz: Sure, I just removed all changes.

nnoll:
My main concern is this would increase our already busy plot. Can we have it off by default but include the option?

shijurodhaz:
> My main concern is this would increase our already busy plot. Can we have it off by default but include the option?

@nnoll Sure, I just made those changes. It is now off by default and can be toggled when the legend is clicked.

rneher:
Cumulative cases here are calculated as `1-susceptible`. This is essentially the same as `recovered` plus `deaths`, but without the delay. Given that `deaths` are about 1% of cases, the new curve is almost the same as a shifted version of `recovered`. 

People should probably only show up as cases once they are infectious. Hence this curve is really shifted forward too much and should rather be the integral of people entering the infectious category. So I am not convinced this adds a lot.

nnoll:
I agree with @rneher. For this to be useful, it should be accurate. Which means either we add cumulative to the user result category and just compute directly from the simulation, or one computes this directly from differences between infectious and exposed in subsequent time points.

bapngo1122:
> ## Related issues and PRs
> * Related to issue #517
> 
> ## Description
> Add a new line to the results plot to show the cumulative cases in the simulation. This is calculated using population - current susceptible
> 
> ## Impacted Areas in the application
> src/components/Main/Results/DeterministicLinePlot.tsx
> src/components/Main/Results/ChartCommon.ts
> 
> ## Testing
> Check the results plot and compare with population/ susceptibility values

bapngo1122:
Bắp

bapngo1122:
Duplicate of #

Issue #612, opened 2020-04-20
IE11: Two crashes


gj262:
# 🐛 Bug Report

run() crashes with "Worker Error" and "Promise is undefined".

Changing the scenario results in an error page. Something to do with immer.

## How to reproduce

IE11 and click Run
IE11 and select a scenario

![0003screenshot (1)](https://user-images.githubusercontent.com/6854428/79809605-c88e0280-8325-11ea-9e8d-06ec030af9c3.png)

![0003screenshot (1)](https://user-images.githubusercontent.com/6854428/79809664-f1ae9300-8325-11ea-81e4-03a4679175a7.png)

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Browser          | IE11 |
| Operating System | Win 10 |
|                  |

## Related

 -

ivan-aksamentov:
I tried to solve it in 7e56a2ea5b4131a5056bbd76098431aea83ac1ee. 

No idea why immer is involved at all, but I replaced its init to ES5 version. Also added polyfills to the worker, similar to how it's done for the entry point.

Worker still does not run occasionally, now silenty. Not sure what's going on.

abrie:
Has 59f8c85d3 fixed the issue?

gj262:
Yes, both issues are resolved in the latest master.

PR #622, opened 2020-04-21
Remove unncessary console.error

abrie:
## Related issues and PRs

#612 reveals the issue, and #505 is the original implementation.   

## Description

There is an unnecessary console.error() log in the webworker. This PR removes the extra call.

Webworker error logging. 

## Testing

Difficult to test without an error, but #612 provides and indirect example of the behavior.

### DIFF
--- a/src/workers/algorithm/index.ts
+++ b/src/workers/algorithm/index.ts
@@ -20,7 +20,6 @@ export async function run(...args) {
 
     worker.addEventListener('error', (error) => {
       reject(error)
-      console.error('Worker error:', error)
     })
 
     worker.postMessage(args)


src/workers/algorithm/index.ts
```diff
@@ -20,7 +20,6 @@ export async function run(...args) {
 
     worker.addEventListener('error', (error) => {
       reject(error)
-      console.error('Worker error:', error)
```
abrie, line 23: This console.error is not necessary, because the reject() is handled in `Main.tsx`:
https://github.com/neherlab/covid19_scenarios/blob/master/src/components/Main/Main.tsx#L60-L64

PR #637, opened 2020-04-27
Parser, folder and sources for japan data included

pvillamartin:
## Related issues and PRs

 -


nnoll:
This looks great. I fixed the merge conflict and it all runs on my end so I'm happy to merge this. Before  I do, I was wondering if you had/were willing to track down population data for the japanese regions, e.g. JP-Kyoto? If so, we could generate default scenario listings in our menu that will try to estimate some basic parameters for each scenario. Thanks for your contribution!

pvillamartin:
Thank you very much for your work! Perfect, I'll try to track the population data next week and add it to the parser. I'll let you know as soon as I have it. Thank you very much.

pvillamartin:
Hello, I have just added the population data for each prefecture (excepting icu beds, that I wasn't able to find). Also, due to a change of format of the source of patients data, I had to change the parser a little bit. Sadly, they do not provide the number of patients in the hospital and icu for each prefecture. If you need that I add or change anything else, just let me know.

rneher:
Thanks. Looking good, but I have a few change requests:

We use three letter codes as country prefix in `populationData.tsv`. Could you change from two-letter to three letter?

Could you make sure the parser exports to integers, not floats? We also leave missing data as simply empty fields rather than putting `nan`. 

I found some articles on ICU numbers in Japan: 
https://asia.nikkei.com/Spotlight/Coronavirus/Japan-ranks-below-Italy-and-Spain-in-ICU-bed-capacity
If numbers per prefecture are missing, you could just scale the national numbers down. This won't be super accurate, but ok as a starting point.

ivan-aksamentov:
Hi @pvillamartin , please let us know if you are available and could implement the changes that Richard requested. Japan's COVID-19 scenario is very interesting to explore and it would be nice to have the detailed data for it!

Summary of requested changes:
 - [ ] 3-letter coutry codes in `populationData.tsv` instead of 2-letter (for consistency)
 - [ ] parser exports to integers, not floats (will not pass validation and will crash the app)
 - [ ] leave missing data as empty fields rather than putting `nan` (will not pass validation and will crash the app)
 - [ ] add at least basic estimation of hospital and ICU bed capacity, for the entire Japan and for every region (for example, nation-wide scaled down by region's population)

Thanks a lot!

pvillamartin:
Hello, sorry for the late reply. Yes, I am available for implementing the requested changes. I will start right now and let you know as soon as I have it. Thank you very much for your work and patience.

pvillamartin:
Hello, I just implemented the requested changes. I saw that populationData.tst already included ICU beds for Japan following the reference https://doi.org/10.1097/CCM.0000000000004222. For consistency I considered such reference. If you prefer that I use the one you proposed, just let me know.

PR #657, opened 2020-05-07
Ensure severity table is restored from URL

ivan-aksamentov:
## Related issues and PRs

 - 

## Description

This makes sure that the severity table is restored from the shared URL.
The issue was that the severity data was deserialized, but ignored and default data was loaded instead.

UI, initial state

### DIFF
--- a/src/components/Main/HandleInitialState.tsx
+++ b/src/components/Main/HandleInitialState.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react'
+import React, { useEffect, useState } from 'react'
 
 import { isEqual, isEmpty } from 'lodash'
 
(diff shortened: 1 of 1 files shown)


src/components/Main/HandleInitialState.tsx
```diff
@@ -69,8 +77,8 @@ function HandleInitialState({ location, push, component: Component }: HandleInit
     <Component
       initialState={{
         scenarioState,
-        severityName: DEFAULT_SEVERITY_DISTRIBUTION,
-        severityTable: severityDistribution,
+        severityName,
+        severityTable,
         isDefault: isEqual(scenarioState, defaultScenarioState),
       }}
```
ivan-aksamentov, line 83: This is where the default severity data was loaded unconditionally. Now replaced with the deserialized data.

PR #685, opened 2020-05-18
Adding comments for better understanding of code

sravanroy:
My first contribution to this project and looking forward to making more useful contributions in the future!

### DIFF
--- a/data/scripts/validateAge.py
+++ b/data/scripts/validateAge.py
@@ -1,20 +1,28 @@
+## loading the required libraries
 import os, json, yaml
 
 from jsonschema import validate, FormatChecker
 
+## reading path for the age distribution JSON file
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
@sravanroy Hi Sravan!

Thanks, but, honestly, I don't see much value in these comments. They are just stating the obvious facts, trivially deducible from the corresponding code fragment itself. The cost of reading the comment is the same as the cost of reading the code, so it's just a duplication. And duplication should be removed.

Also, this particular code is absolutely not important. This just a one-time converter/sorter script. I am not even sure if it's a part of the pipeline or whether it is used at all. The documentation efforts should be applied elsewhere.If you have time, you could take a look at case data parsers (raw-to-tsv) and converters (tsv-to-json). Parsers for sure, they can use some love.

sravanroy:
Thanks for the feedback!

Issue #688, opened 2020-05-19
Add progress indicator for the results section


ivan-aksamentov:
# 🙋 Feature Request

## 🔦 Context

Currently, when algorithm is running, the progress is only indicated on the "Run/Refresh" button. We need to add some sort of a progress indicator in the results section itself, to make it more intuitive.

## 😯 Describe the feature

 - [ ] should clearly show that the algorithm is running and that the results are about to change
 - [ ] should not obscure the plot completely, such that the moment when changes arrive can be seen
 - [ ] the indicator timing should be fully aligned with the indicator on the button

Various creative solutions should be tried (spinners, text, fade overlay etc.) until good UX is achieved.

## Related

 -
 -

Chong-anson:
Hi, I am trying to make my first commit to an open source project and I would love to claim this issue!

ivan-aksamentov:
@Chong-anson sure, give it a try. Thanks!

Chong-anson:
I suppose we need to add loading progress to cover the plot?

Chong-anson:
Hi Ivan, my idea is to create a component similar to RunButtonContent, which will display a loading spinner when the plot is loading. I'm also thinking adding a props to the ResultCard to "blur out" the plot when it is loading. What do you think about this?

ivan-aksamentov:
Resolved in #706

PR #706, opened 2020-05-27
add progress indicator to the result section

Chong-anson:
## Related issues and PRs

 - add progress indicator to the result section #688 

## Description

 - create a spinner when the plot is loading. The spinner is aligned with the indicator to the button.

 - add PlotSpinner component which renders the spinner logo. 
 - add ternary logic to the plot components so that when the result "isRunning", it will render the spinner icon. Otherwise, it will the normal components
 - edit result css file to centre the spinner icon

### DIFF
--- a/src/components/Main/Results/PlotSpinner.tsx
+++ b/src/components/Main/Results/PlotSpinner.tsx
@@ -1,0 +1,48 @@
+import React from 'react'
+
+import i18next from 'i18next'
+import Loader from 'react-loader-spinner'
+
+export interface RunningSpinerProps {
+  size: number
+}
+
+export function RunningSpinner({ size }: RunningSpinerProps) {
+  return <Loader type="ThreeDots" color="#495057" height={size} width={size} />
+}
+
+export interface PlotState {
+  Icon: React.ElementType
+  text: string
+}
+
+const states: Record<'normalRunning' | 'autorunRunning', PlotState> = {
+  normalRunning: {
+    Icon: RunningSpinner,
+    text: i18next.t(`Running...`),
+  },
+  autorunRunning: {
+    Icon: RunningSpinner,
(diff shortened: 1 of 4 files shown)


ivan-aksamentov:
@Chong-anson Thanks, it looks promising, however I insist on this important feature (from the original issue):

>  - should not obscure the plot completely, such that the moment when changes arrive can be seen

The whole idea of the app is to allow user to explore different parameters and to see how they affect the results (for example, when relaxing the quarantine measures, death plot goes up, etc.) 
In this PR, when loading indicator is shown, the plot disappears and the entire results section changes its layout. This makes it difficult to track the changes visually when changing parameters. 

Instead of conditional rendering (spinner vs plot) we could implement a semi-transparent overlay on top of the plot, such that the data and the transition from one result to another is visible at all times.

Note, that rendering the whole plot from scratch is very costly (takes from 500 to 1000ms on average machine), so we want to ensure that only the changed items are re-rendered (for example, axes, grid and legend stays the same). This means that we need to avoid unmounting the plot component, changing it's size and changing its props unnecessarily.

Also, currently, the size is changing when switching between the spinner and the plot. Implementing overlay should hopefully fix that. Additional bonus for ensuring  that there is no size jump on first render (you know, when the results section "unrolls" down on app load).

This is a creative issue, so feel free to experiment, implement your own idea of how it may look like and refactor the code structure when needed.

Chong-anson:
Thanks for your detailed comment! I will work on that!

Chong-anson:
Hi Ivan, I edited the components in ResultCard and added some CSS. Sorry for the delay, my internet was down. Currently working on the "unrolling issue" in the first render.

ivan-aksamentov:
This looks good thanks!
Resolves #688

Issue #715, opened 2020-05-30
Option for adding Google Mobility data as an additive mitigation NPI


osnofas:
# 🙋 Feature Request
I'd like an option to quickly add Google Mobility data (https://www.google.com/covid19/mobility/) difference from baseline as a daily mitigation intervention.

## 🔦 Context
I'm trying to visually "fit" the model's predicted cases curve to the observed cases by fiddling with the assumptions of the epi parameters and the mitigation interventions. For a given set of epi parameters (mostly R0), making the predictions consistent with the observed case numbers (which are displayed by the tool) and other locally available data not displayed by the tool (such as ICU usage) is hard as it requires varying the mitigation intensity week by week. In regression analysis, mobility data captures most of the effect of the mitigation strategies such as the different lockdowns (https://papers.ssrn.com/sol3/papers.cfm?abstract_id=3581633).

Using the "mitigations" panel to "draw" manually what would quickly and more precisely captured by mobility data is not optimal.

## 😯 Describe the feature
A tick box in the mitigations panel would additively add Google Mobility data for the selected geographical area as an intervention. The subindices for workplaces and transit should be used rather than the other subindices that include things such as parks, given that we know that transmission is most likely to occur indoors and in sustained interactions.

Additively: the rest of the panel would work as today, but allowing for negative numbers. If the google mobility data for a given day is -80% and interventions for that day add to a -20% in the panel (which currently displays it's sign inverted for clarity purposes), the sum of the interventions would yield -80% -(-20%) = 60%. This would allow for further adjustments.

But the panel should mostly keep working as today.

Thanks!

rneher:
adding mobility data could be fun. We could overlay this with the Rt graph. But I haven't looked into Google's API and how we'd match up keys for different regions.

osnofas:
The csv file is really simple

https://www.google.com/covid19/mobility/

CSV: https://www.gstatic.com/covid19/mobility/Global_Mobility_Report.csv?cachebust=6d352e35dcffafce
countries are ID'd by name and by 2-digit ISO country code. Country totals have missing data on other columns that otherwise indicate geographical region (state/provinces, county/municipality/department).

Here's the csv documentation:
https://www.google.com/covid19/mobility/data_documentation.html?hl=en

Thanks!

TakumaKira:
Hi there!

I tried adding this feature. I added "ADD MOBILITY DATA" and "REMOVE MOBILITY DATA" buttons on the MITIGATION panel.
[...]

PR #689, opened 2020-05-20
Add a simple CLI for use in unified model interface

adityasharad:
## Related issues and PRs

 - Related to https://github.com/neherlab/covid19_scenarios/issues/663, though not yet a fully-featured CLI.

## Description

This PR adds a simple CLI entry point that:
- reads input JSON (of type `ScenarioData`) from a specified input path
- runs the model by calling into `run.ts`
- writes output JSON (of type `AlgorithmResult`) into the specified output path

This allows the model to be run in a unified model UI, such as the one we are maintaining at https://github.com/covid-modeling on the GitHub side. We are already using a version of this at https://github.com/covid-modeling/model-runner to run this model, but are currently tied to my fork that has the CLI entry point and an older version of your model. Having this code as part of the model repo enables the unified interface to keep up with model changes, and so better represent the latest state of your model.

@ivan-aksamentov @nnoll I would be grateful for a review, and hope this will minimally impact your existing work. Let me know if you have questions.

## Impacted Areas in the application

No changes to existing code, introduces new entry point that calls `run`.

## Testing

1. Usual schema preparation: `yarn schema:clean && yarn schema:totypes`.
1. Write a JSON object of type `ScenarioData` in a file, say `input-file.json`. One of the existing objects in `scenarios.json` will work.
1.
    ```
    yarn babel-node --extensions ".ts" src/algorithms/cli.ts path/to/input-file.json path/to/output-file.json
    ```
1. This will create output data at `path/to/output-file.json`, which is assumed to not already exist.

Issue #663, opened 2020-05-08
Introduce NPM package


ivan-aksamentov:
# 🙋 Feature Request

## 🔦 Context

We would like to allow our users to run the algorithm "in bulk", for different scenarios, without the need to operate the UI. 

## 😯 Describe the feature

NPM package would contain all the necessary algorithmic parts, data and schemas and will implement:

 - [ ] CLI
 - [ ] programmatic Node.js API (being able to be imported)
 - [ ] in Typescript, with typings included
 - [ ] reading input parameters from JSON file
 - [ ] overriding input parameters with command line arguments
 - [ ] producing TSV outputs (detailed or summary)
 - [ ] using schemas as much as possible: to generate command line arguments, serialize and validate payloads etc.
 - [ ] (possibly) structure the package such that it would allow to easy integrate it into API servers and other value-added products
 - [ ] (possibly) wrappers for different languages (Python and R are the most requested)

## Related

 -
 -

ivan-aksamentov:
#689 add some very basic CLI.

Some immediate improvements would be:

 - [ ] create a `cli/` directory
 - [ ] implement a proper CLI arguments parser (with a library)
 - [ ] wrap `run()` into another function, which can be exported from an NPM package for programmatic access. `run` function should not be exposed, to reserve the right of changing it's signature internally.
 - [ ] validate user's inputs using a schema
 - [ ] severity and age distribution json files are imported (bundled along with corresponding `get*()` function), but should be read as JSON files instead
 - [ ] probably no need to  validate severity and age distribution separately, because they are validated upon generation. The validation in `get*()` functions in the app is mostly a precaution for development mode
 - [ ] the external API should use `Shareable` schema and corresponding deserialization and validation (reuse the code for URL and scenario imports). This however assumes user provides their own severity and age distribution. One can still read am merge the default ones if needed though.
 - [ ] reading JSON files could be done with `fs-extra`, instead of a manual solution
 - [ ] for console output the `console.log` should not be used outside of debugging. Use more fine-grained `console.{info,warn,error}`, or, better, a proper solution such as `debug` package.
 - [ ] CLI and programmatic API should be transpiled with `rollup` and published on NPM

PR #725, opened 2020-06-06
Command line runs

bettbra:
Add a yarn target for running the model via the command line.

## Description
These changes allow the model to be run via the command line via a `yarn cli` call. Two arguments are required: `--scenario` and `--out`.
 
The changes are all pretty trivial. Some command line parsing is done via `yargs` in `src/algorithms/cli.ts` (and so a dependency is introduced on the [yargs](http://yargs.js.org/) package in `package.json`).

## Testing
1. Copy the example scenario parameters shown on `docs/comand_line.md` into a file `scenario.json`.
1. At the command line, execute `yarn cli --help` to see the usage message.
1. At the command line, execute `yarn cli --scenario=scenario.json --out=output.json`.
1. Confirm that `output.json` was created by the run and has reasonable-looking output.

## Additional Notes
* We are using these mods as we do some grid searching across input parameters.
* I recognize this is our first pull request to your project. We hope we have adhered to the contribution guidelines, but of course let us know how we can modify the changes to better suit the project. And equally, zero worries if you just want to reject the pull request.
* Thank you for your hard work on this project and for making it public so that we can all benefit and contribute.

### DIFF
--- a/src/algorithms/cli.ts
+++ b/src/algorithms/cli.ts
@@ -1,5 +1,7 @@
 /* eslint-disable no-console, unicorn/no-process-exit */
 import * as fs from 'fs'
+import * as yargs from 'yargs'
+
 import { run } from './run'
 import { ScenarioFlat, ScenarioData } from './types/Param.types'
 import { getAgeDistribution } from '../components/Main/state/getAgeDistribution'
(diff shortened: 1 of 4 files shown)


ivan-aksamentov:
@bettbra Hi Bradley,
Thanks! I should probably note that at this point we prioritize the app and we don't officially support the CLI. Personally, I never tried to run it and don't know if it runs correctly. Theoretically it should :) That being said, it wouldn't hurt to have a CLI if community wants it.

However, before merging, I need to ask the folks at GitHub/Microsoft who originally contributed it in #689 . @adityasharad Could you please review the changes and see if it is okay for your use-case? I'll be happy to merge then.

@bettbra Could you please tell us more about your project, so that we could take this information into account when planning the improvements of the algorithm?

Any additional improvements are welcome, including any of the items listed in https://github.com/neherlab/covid19_scenarios/issues/663

bettbra:
Hi Ivan,

Thanks for your reply. Right, I should have checked to try to infer if another team was also using the CLI. I suspect these changes here won't work for them (e.g., @adityasharad) as they already have dependencies on the existing form, but maybe we can coordinate. For us it's convenient to be able to control all parameters---scenario, age, severity---at the command line, meaning not only the first set. So really the change here is focused on adding the last two (age, severity) and using yargs for argument parsing purely for convenience.

I also see a bunch of eslint warnings that I'll fix; forgive me, didn't realize those were getting run. I'll eliminate all the warnings to adhere to the project conventions.

As for what we're using the project for, we're trying to vary R_0 using mobility data, very similar in spirit to #715. We execute large numbers of runs as we assess sensitivities to various input parameters, doing this on the command line in a large gridded compute environment. The UI is fantastic and we use it for quick sanity checks, but our heavy lifting is done via the CLI.

Apologies for the long reply. If the community would favor making the CLI a supported interface we'd be enthusiastic supporters and commit to helping maintain it. Thanks again @ivan-aksamentov. I'll await additional overall feedback and at least fix those warnings as mentioned.

 --Brad

ivan-aksamentov:
@bettbra 

> For us it's convenient to be able to control all parameters---scenario, age, severity---at the command line

Yes. That is probably the most useful interface for the community. 

Feel free to propose an interface that is useful for you. For example we could every parameter to be exposed from both, files and command line, and command line would take precedence. This way one could have a base config and vary a given parameter. We may also read json from stdin, so that users could pipe the autogenerated configs through a shell.

> I also see a bunch of eslint warnings

Note that the current eslint config is mostly for the app, so some of these warning may not necessarily make much sense for a Node script. In this case, feel free to either (1) add an eslint-ignore comment for the file or (2) add an item in the `override` section of the eslint config file or (3) `.eslintignore` the cli file/directory and create a separate config. Whichever suits your needs the best.

> As for what we're using the project

Wow, that sounds cool! Do you have a link? 

>  helping maintain it

That would be fantastic!  Feel free to pick any if the items in https://github.com/neherlab/covid19_scenarios/issues/663 and add yours too. We are looking forward to any feedback and contributions.

I guess currently you have to pull the project from GitHub in order to use it, maybe as a submodule or similar. Would it be useful if we create and NPM package with a CLI and programmatic Node.js interface? We could automate the build and publishing as well. One advantage is more convenient versioning and reduction of breakage, because we are currently not tracking how our changes affect the CLI at all.

src/algorithms/cli.ts
```diff
@@ -13,44 +15,88 @@ const handleRejection: NodeJS.UnhandledRejectionListener = (err) => {
...
-    usage()
+/**
+ * Get severity distribution data. If a file is specified on the command
+ * line, give priority to its contents, else load a default distribution.
+ *
+ * @param inputFilename The path to the file.
+ */
+function getSeverity(inputFilename: string | undefined) {
+  if (inputFilename) {
+    const data = readJsonFromFile<any>(inputFilename)
```
adityasharad, line 38: I think you can type this as `SeverityDistributionData` instead of `any`. That type can be added to the import from `./types/Param.types`.

src/algorithms/cli.ts
```diff
@@ -13,44 +15,88 @@ const handleRejection: NodeJS.UnhandledRejectionListener = (err) => {
...
+ * line, give priority to its contents, else load the distribution
+ * name as specified in the scenario parameters.
+ *
+ * @param inputFilename The path to the file.
+ * @param name The age distribution name to use if no file is
+ *             specified.
+ */
+function getAge(inputFilename: string | undefined, name: string) {
+  if (inputFilename) {
+    const data = readJsonFromFile<any>(inputFilename)
```
adityasharad, line 58: I think you can type this as `AgeDistributionData` instead of `any`. That type can be added to the import from `./types/Param.types`.

rneher:
this does look like a very useful direction and we might consider ditching our parallel implementation in python for model fitting.

ivan-aksamentov:
@adityasharad Thanks for reviewing.

@adityasharad @bettbra  I will merge and then fix the kinks on master. Let me know if there's anything to ammend or what other improvements you have in mind.

bettbra:
Thanks all for the input. I fixed the eslint kinks on my branch without changing the eslint settings. I can open another pull request for it if you'd like; apologies for the delay. Thanks @adityasharad and @ivan-aksamentov for your feedback.

bettbra:
Oh, and I should have commented @rneher: We're grateful for any direction you want to take the project that allows command line running, or more broadly programmatic modification of parameters. If this does the trick for the community, super, but happy to switch to any other language or methodology.

Thanks again for all your hard work on this.

PR #731, opened 2020-06-09
Feat/weekly deaths

rneher:
## Description
As we progress further into the pandemic, total deaths are not very useful to assess current trends. This PR adds weekly deaths to the graph and also changes the interval for cases to weekly. This has the benefit of smoothing out weekday variation. Need to adjust colors. 

plot and plotdata preparation. 

## Testing
do we like it?

### DIFF
--- a/src/algorithms/preparePlotData.ts
+++ b/src/algorithms/preparePlotData.ts
@@ -15,6 +15,7 @@ export function preparePlotData(trajectory: Trajectory): PlotDatum[] {
       overflow: verifyPositive(x.current.overflow.total),
       recovered: verifyPositive(x.cumulative.recovered.total),
       fatality: verifyPositive(x.cumulative.fatality.total),
+      weeklyFatality: verifyPositive(x.cumulative.fatality.total - middle[i > 6 ? i - 7 : 0].cumulative.fatality.total),
     },
     // Error bars
     areas: {
@@ -45,6 +46,13 @@ export function preparePlotData(trajectory: Trajectory): PlotDatum[] {
       fatality: verifyTuple(
         [verifyPositive(lower[i].cumulative.fatality.total), verifyPositive(upper[i].cumulative.fatality.total)],
         x.cumulative.fatality.total,
+      ),
+      weeklyFatality: verifyTuple(
+        [
+          verifyPositive(lower[i].cumulative.fatality.total - middle[i > 6 ? i - 7 : 0].cumulative.fatality.total),
+          verifyPositive(upper[i].cumulative.fatality.total - middle[i > 6 ? i - 7 : 0].cumulative.fatality.total),
+        ],
+        x.cumulative.fatality.total - middle[i > 6 ? i - 7 : 0].cumulative.fatality.total,
       ),
     },
   }))
--- a/src/components/Main/Results/ChartCommon.ts
+++ b/src/components/Main/Results/ChartCommon.ts
@@ -116,14 +116,14 @@ export const areasToPlot: LineProps[] = [
   },
 ]
 
-export function observationsToPlot(casesDelta: number): LineProps[] {
+export function observationsToPlot(): LineProps[] {
   return [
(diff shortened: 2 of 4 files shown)


ivan-aksamentov:
@rneher Yes, this indeed gives some better feeling for the trend. 

Model's weeklies don't display uncertainty. Bug or feature?
Also, the tooltip text casing in inconsistent.

src/algorithms/preparePlotData.ts
```diff
@@ -15,6 +15,7 @@ export function preparePlotData(trajectory: Trajectory): PlotDatum[] {
       overflow: verifyPositive(x.current.overflow.total),
       recovered: verifyPositive(x.cumulative.recovered.total),
       fatality: verifyPositive(x.cumulative.fatality.total),
+      weeklyFatality: verifyPositive(x.cumulative.fatality.total - middle[i > 6 ? i - 7 : 0].cumulative.fatality.total),
```
ivan-aksamentov, line 18: @rneher 
If I read it correctly, the conditionals `i > 6 ? i - 7 : 0`, or even the entire subtraction expressions seem to be duplicated. Could we extract them into a variable and bubble it up ? For the sake of self-documentation, some good name for this variable would be nice to have as well. I'm not sure if I correctly understand what this expression calculates and why.

This refactoring will probably require introducing a function scope with `return` statement and all this code will be indented to the right, so we may do this refactoring afterwards, to keep the current diff clear, as it is.

ivan-aksamentov:
Patched the text casing.

src/algorithms/preparePlotData.ts
rneher: we should call it `day`! and yes, its duplicated. But is hard to see how the function call is any shorter than the arguments. Maybe adding a comment is good enough...

rneher:
the missing uncertainty is a bug... but getting this uncertainty correctly is actually a little tricky algorithmically.

src/algorithms/preparePlotData.ts
ivan-aksamentov: I never said "shorter"! All I care is the programmer's performance (code that is easy to understand) and runtime performance (avoiding unnecessary computation).

Not function call, just a variable:

```js
something.map((...) => {
   const day = i > 6 ? i - 7 : 0
   const thingy = something - middle[day].cumulative.fatality.total

   return {
     one: verify(a - thingy),
     two: verify(b - thingy),
     three: verify(c - thingy),
  }
}
```

or better (so that the usage of the variable is closer to the declaration):

```js
something.map((...) => {
   const day = i > 6 ? i - 7 : 0
   const thingy = middle[day].cumulative.fatality.total

   const one = verify(a - thingy)
   const two = verify(b - thingy)
   const three = verify(c - thingy)

   return { one, two, three }
}
```

also maybe

```js

  const one = a - thingy
  const two = b - thingy
  const three = c - thingy

  return ensureSomeParticularPlotDataRequirements({ one, two, three })

```
to cut the duplication further.

Anyways, refactoring is out of scope of this PR I think. The clear diff is more important here, to see the changes. We can sit down and refactor later.

ivan-aksamentov:
Should we also include the number to the results and emit it into the exports? It looks somewhat easy to derive right now, so maybe not. Bunt if the algorithm for uncertainty will be tricky, we may want to include it, for consistency.

src/algorithms/preparePlotData.ts
rneher: I vaguely remember:
`#define SEVEN_DAYS_AGO(d) ( d>6 ? d-7 : 0)`

rneher:
> Should we also include the number to the results and emit it into the exports? It looks somewhat easy to derive right now, so maybe not. But if the algorithm for uncertainty will be tricky, we may want to include it, for consistency.

might be a good idea. The issue with the upper and lower is that weekly increases of the lower cumulative death curve might be higher than the weekly increase of the upper, thereby screwing up what is min and max. so maybe we just have to bite the bullet and compute this along with the result.

rneher:
I fixed some of these issues and made a separate list of plots that should be enabled by default.

PR #757, opened 2020-07-17
added new palliative flux

tiagoandresvaz:
## Related issues and PRs

#703
#753

## Description

To implement the new proposed flux in the model and interfaces according to the discussion above.

Age-Specific Parameters, and Model.

## Testing

Edit age-specific parameters to see changes in deaths modeled in results.


rneher:
Hi @tiagoandresvaz,

this now looks very good to me. I think this addition is needed. Before we can merge this, we need to address a few things

 * how do we handle old jsons. filling in `palliative=0` reproduces the old behavior, so that would be an option. Currently the app with old jsons, but all values of the computation past infectious are `undefined` since undefined values of `palliative` enter. 
 * add the new addition to the figure on the about page and add a paragraph to the `updates` page
 * fix a few little things here and there (linting etc).

I am also wondering whether we should rename `critical` to `ICU` in the table. Now that we also have `palliative`, there are critically ill people that don't end up in ICU. But I am not sure what the least confusing terminology is... (@ivan-aksamentov ?? )

richard

ivan-aksamentov:
> whether we should rename critical to ICU in the table

Does it relate to ICU on other places? ICU beds, ICU on the plot? Should rename in all places then?
And also, in export/import files etc.

ivan-aksamentov:
@tiagoandresvaz @rneher 

> how do we handle old jsons

Ideally, the workflow is as follows:

 - [ ] increment schema version (to 2.1.0):
https://github.com/neherlab/covid19_scenarios/blob/f62c8dc9bc3b9dc6213b67c4379ab1d218d6b6ad/schemas/_SchemaVer.yml
  - [ ] add serialization and deserialization path for the new version
     All related code is [here](https://github.com/neherlab/covid19_scenarios/tree/fa2057e23e4ddac90c77bc3cffd88d9852386e24/src/io/serialization
).   For example, the choice of deserializer (depending on `schemaVer` of the incoming object) happens [here](https://github.com/neherlab/covid19_scenarios/blob/f62c8dc9bc3b9dc6213b67c4379ab1d218d6b6ad/src/io/serialization/deserialize.ts#L20). When a JSON or URL blob containing  `schemaVer` property arrives there, it will be directed to an appropriate code path.

 - [ ] generate and commit to git the generated typings and validators from the old schema version (2.0.0).
   Go to master branch (before schema modification), generate typings and serializer files (`yarn schema:totypes`), save directory `src/.generated/latest` to temporary place, go to branch of this PR, move the saved directory as `src/.generated/2.0.0`, commit it. This way we can have `src/.generated/latest` containing validators for `2.1.0` and `src/.generated/2.0.0` for `2.0.0`.

 - [ ] use these typings and validators to modify the deserialization code for the old version (2.0.0): mostly adjust imports and fill `palliative` with `0` on output. Note that we don't have to implement serializer, because we always save using to the latest schema version. 

This should automatically ensure that the application code always receives the object of version 2.1.0, however users with old JSONs and URLs will still able to load their data.

ivan-aksamentov:
> fix a few little things here and there (linting etc)

These are mostly formatting warning, can be fixed with `yarn format:fix`.
Also, `yarn lint` to run the entire static analysis stack.

tiagoandresvaz:
@ivan-aksamentov the app is running nicely after modifications that you and @rneher recommended.
But when loading jsons created with schema 2.0.0 version, it is rising: expected `schemaVer` to be one of `[2.1.0]`, but received: 2.0.0

it seems that only one version is allowed to be received. Can you confirm where is this configuration?

src/io/serialization/v2.0.0/serialize.ts
```diff
@@ -1,35 +1,24 @@
-import { trim, isEqual } from 'lodash'
+/* eslint-disable camelcase */
+import { trim } from 'lodash'
 
 import Ajv from 'ajv'
 import ajvLocalizers from 'ajv-i18n'
 
-import validateShareable, { errors } from '../../../.generated/latest/validateShareable'
+import validateShareable, { errors } from '../../../.generated/2.0.0/validateShareable'
+import { Convert } from '../../../.generated/2.0.0/types'
```
ivan-aksamentov, line 8: Serializer for the old version should use the types, converters and validators from the old version of schema.

src/io/serialization/v2.0.0/serialize.ts
```diff
@@ -64,43 +53,24 @@ function convert(shareableDangerous: Record<string, unknown>): Shareable {
...
+      ...shareable.severityDistributionData,
+      data: shareable.severityDistributionData.data.map((severity) => ({ ...severity, palliative: 0 })),
     },
-    ageDistributionData,
-    severityDistributionData,
   }
+
+  // Delegate to the next version of deserializer
+  return v2_1_0[schemaVerNext].deserialize(JSON.stringify(shareableNew))
 }
```
ivan-aksamentov, line 74: Deserialization function for v2.0.0 deserializes the received object of v2.0.0 as before, but then also adjusts the object such that it looks like an object of v2.1.0, then passes it to the deserializer of v2.1.0.

It's a bit redundant, as most of the checks will run twice. But it's easy to implement and maintain.

src/io/serialization/versioning.ts
```diff
@@ -21,6 +22,7 @@ export interface Serializers {
 export const SERIALIZERS = new Map<string, Serializers>(
   Object.entries({
     ...v2_0_0,
+    ...v2_1_0,
   }),
 )
```
ivan-aksamentov, line 27: We maintain a list of all serializers, in order to route incoming objects depending on their versions. The serializer v2.0.0 was missing, so the I/O on the objects of v2.0.0 was always failing. This fixes that problem.

src/io/serialization/v2.0.0/serialize.ts
```diff
@@ -1,35 +1,24 @@
...
-  if (process.env.NODE_ENV !== 'production' && !validateShareable(JSON.parse(serialized))) {
-    throw errors
-  }
-
-  return serialized
+export function serialize(_0: ScenarioParameters): string {
+  throw new Error(`: Developer error: This function should never be called.
+  Serialization must always use the latest version of the serializer, but serializer of version "${schemaVer}" is called instead.
+  Change the callee to use the latest version.`)
 }
```
ivan-aksamentov, line 22: Now that we have serializer of v2.1.0, the old serializer (v2.0.0) should never be called, as we always want to output files and URLs in the new format.

src/components/Main/validation/schema.ts
```diff
@@ -156,11 +156,15 @@ export const schema: yup.Schema<FormData> = yup
...
+            palliative: percentageSchema.required(MSG_REQUIRED),
             critical: percentageSchema.required(MSG_REQUIRED),
             fatal: percentageSchema.required(MSG_REQUIRED),
             isolated: percentageSchema.required(MSG_REQUIRED),
           })
-          .required(MSG_REQUIRED),
+          .required(MSG_REQUIRED)
+          .test('max', MSG_EXCEED_100, ({ critical, palliative }: SeverityDistributionDatum) => {
+            return critical + palliative <= 100
+          }),
```
ivan-aksamentov, line 167: Moved the form validation test to the object scope, so that it is easier to understand and type properly.

src/components/Main/SeverityTable/SeverityTable.tsx
```diff
@@ -114,7 +116,10 @@ export function getErrorMessages(t: TFunction, errors?: FormikErrors<AgeGroupRow
...
     const ageGroup = Object.values(AgeGroup)[rowNum]
 
-    if (err) {
+    if (typeof err === 'string') {
+      return [...result, t(`Error in row "{{ageGroup}}": {{message}}`, { ageGroup, message: err })]
+    }
+    if (typeof err === 'object') {
       const messages = Object.entries(err).map(([column, message]) =>
         t('Error in column "{{column}}", row "{{ageGroup}}": {{message}}', { column, ageGroup, message }),
       )
```
ivan-aksamentov, line 125: Adjusted the errors display logic in severity table: when the new check in yup schema fails, the result is being sent as an array with a single string in it (because the check is attached to a particular object in an array of objects).

ivan-aksamentov:
@tiagoandresvaz I added a few changes and commented on them above. This should fix the remaining problems.

Note that I also removed the `src/.generated/latest` from version control, as the latest types are a subject to change and should always be generated on the fly to avoid mistakes. By contrast, the old types (v2.0.0) will never change anymore, so they can be committed.

tiagoandresvaz:
@ivan-aksamentov 

> Please pull and check that this stuff works on your side.

Pull done! It is ok on my side!

ivan-aksamentov:
@tiagoandresvaz okay, let's merge this!

PR #760, opened 2020-07-23
fix: adjust Brazilian parser for the new upstream file format

### LINKED ISSUE #345: Brazil regions for populationData.tsv
We have the country Brazil now, but the smaller divisions are missing. We have case data for them already. Maybe @gstvribs can help?

ivan-aksamentov:
- Resolves #345 

brazil.io updated their URLs and they now compress the files.
This PR adjusts our parser for that.

Sorry for the large diff, my text editor formats files automatically and I did not bother to undo.

### DIFF
--- a/data/parsers/brazil.py
+++ b/data/parsers/brazil.py
@@ -1,9 +1,11 @@
-import sys
-import requests
 import csv
+import gzip
 import io
-
+import sys
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


rneher:
thanks. This looks good to me. I suppose we can release this along the pdf prettifyications?

PR #777, opened 2020-07-31
feat: seroprevalence

### LINKED ISSUE #779: Seroprevalence
In order to improve the trajectories further into epidemic we want to account for the existing immune population. This population is no longer susceptible to the disease and "bypasses" the model.

ivan-aksamentov:
Resolves #779
Contributes to #778 

Adds seroprevalence parameter to the:

 - [x] UI (form)
 - [x] algorithm
 - [x] schema (bump to v2.2.0)
 - [x] I/O
 - [x] data generation

seroprevalence


rneher:
I removed the immune from the recovered again, as it messed up the pie chart with outcomes. I now instead have `pop.immune` are an array in the results structure. The algorithm deals with it correctly. We can see whether/how we want to show the this fraction of the population in the plots and tables.

rneher:
the pdf print is broken for me.

ivan-aksamentov:
> the pdf print is broken for me.

Fixed in https://github.com/neherlab/covid19_scenarios/pull/777/commits/8ac2dad5b2f93a602c98678256f0cfa7e9cee16d

rneher:
Hi Ivan,

I worked a little more on the seroprevalence branch. this now has new data (the script still has some small problems -- produces NaN outputs when some data are missing). 

I made the R0 curve extend over the entire range, not just the part that we are simulating now. But this is pretty hacky, but I first wanted to see how this looks:
https://github.com/neherlab/covid19_scenarios/blob/feat/seroprevalence/src/algorithms/run.ts#L69

PR #802, opened 2020-08-27
Update CAN-British Columbia.tsv

hdsheena:
Added hospitalization numbers since Aug 6. If others are interested, I'll go back further. Let me know if I need to add "dummy" data to weekend rows, and how to best do so.

## Related issues and PRs
None - 

## Description
Added hospitalization numbers since Aug 6. If others are interested, I'll go back further. Let me know if I need to add "dummy" data to weekend rows, and how to best do so.

data for BC Canada

## Testing
Load this data (BC) wherever it's used to be sure it works!


ivan-aksamentov:
Hi @hdsheena, thanks!

This .tsv file is auto-generated. We periodically re-download data from our sources
https://github.com/neherlab/covid19_scenarios/blob/master/data/sources.json
and create all these .tsv files from scratch, so your changes will be overwritten on next update unfortunately.

Many countries have dedicated scripts for this. For example, the ,tsv file for Canada is generated using this script:
https://github.com/neherlab/covid19_scenarios/blob/master/data/parsers/canada.py

Do you have a source from where we could pull these hospitalization numbers automatically once in a while? If not, maybe you could create such a source? A git repo with a CSV, JSON or similar file would be perfect. We would them merge the data from the current source with the new one.

Also, it is important that the sources update often themselves. In case you create a new data repo, will you be able to provide periodic updates for it?

PR #864, opened 2021-03-08
added info to release notes

tiagoandresvaz:
add  information to release notes and close the bug as solved.
"The case-count export was restricted to the last 200 days, assuming that most people study recent dynamics. "


rneher:
thanks!

